PR #3, opened 2017-04-08
protocol: use enums where it makes sense

Keruspe:
Just seemed to make sense for those


dvdhrm:
Can we please turn all of them into enums? Lets avoid preprocessor use for the sake of better debugging, and use enums for single bits as well.

Keruspe:
Sure, wasn't sure whether we wanted to do that for bitmasks too or not.

teg:
No need to name these things though? Could all be anonymous.

Keruspe:
Should the flags be anonymous or all of the enums?

dvdhrm:
You can leave all of them anonymous. C does not have real type-safety on those, so we avoided using them so far.

Keruspe:
Should be ok now

PR #4, opened 2017-04-11
match: store destination id instead of unique name

Keruspe:


src/match.c
```diff
@@ -69,7 +72,7 @@ static bool match_rule_keys_match_filter(MatchRuleKeys *keys, MatchFilter *filte
         if (keys->filter.sender && !c_string_equal(keys->filter.sender, filter->sender))
                 return false;
 
-        if (keys->filter.destination && !c_string_equal(keys->filter.destination, filter->destination))
+        if (keys->filter.destination && keys->filter.destination != filter->destination)
```
teg: This won't quite work. :1.0 is a valid unique name. I guess we need to introduce a PEER_ID_INVALID constant (-1) to indicate that the ID is unset. Or something like that.

Keruspe:
Wasn't sure about the id 0 handling.
I went ahead with the PEER_ID_INVALID

teg:
David pointed out that this won't quite work. Firstly, any unique name might be passed in (even if it would never actually be used), so we could easily end up with id -1 (or indeed anything else) in a match rule. Moreover, not only unique names are supported for matching, but also well-known names, so resolving this at install time to the peer id is not going to work I'm afraid. At least it is not obvious how. I guess the XXX comment should just be removed for now.

Keruspe:
Heh, makes sense, closing then

PR #8, opened 2017-05-24
Silence compiler warnings

Keruspe:
At least the first one in launch is relevant as we're using autocleanup for a maybe-unitialized variable (in the case when we return in the first error path)

### DIFF
--- a/src/launch/main.c
+++ b/src/launch/main.c
@@ -649,7 +649,7 @@ exit:
 
 static int manager_load(Manager *manager) {
         const char suffix[] = ".service";
-        _c_cleanup_(c_closedirp) DIR *dir;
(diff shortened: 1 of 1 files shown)


Keruspe:
Done.

Nice work btw, I've been trying it as a session bus for my gnome session for a whole day without any major issue :)

teg:
Thanks! And bug reports welcome of course, I'm not aware of any outstanding issues at the moment (apart from our lack of policy support and accounting, that is).

Keruspe:
Yep, lack of policy is what's preventing me from testing it as the system bus.
Just opened an issue about the only noticeable difference from my session runnig dbus-daemon.

PR #27, opened 2017-08-24
selinux: do logging via audit when available

teg:


src/util/selinux.c
```diff
@@ -326,6 +327,27 @@ int bus_selinux_check_send(BusSELinuxRegistry *registry,
...
+static int bus_selinux_log(int type, const char *fmt, ...) {
+        _c_cleanup_(c_freep) char *message = NULL;
+        va_list ap;
+        int r;
+
+        va_start(ap, fmt);
+        r = vasprintf(&message, fmt, ap);
+        va_end(ap);
+        if (r < 0)
+                return r;
```
dvdhrm: What happens if we return `<0` from log-callbacks?

src/util/selinux.c
```diff
@@ -326,6 +327,27 @@ int bus_selinux_check_send(BusSELinuxRegistry *registry,
...
+
+        /* XXX: we don't have access to any context, so can't find
+         * the right UID to use, follow dbus-daemon(1) and use our
+         * own. */
+        r = util_audit_log(message, getuid());
```
dvdhrm: Why is this a `TODO` item? There is nothing to fix. The callback has no user-data so this should just say that we use our own uid, since selinux does not allow us to pass a callback. Or is there anything we can do about it, which does not end up being a major hack?

teg: We could take it up upstream (to ask them to give us a callback with userdata), not aware of any other work-around.

src/util/selinux.c
teg: I followed the behavior of the default log callback in libselinux, which returns negative error codes on error. However, looking through libselinux, it appears nothing catches these return values.

PR #31, opened 2017-08-24
Add meson build option for selinux

floppym:
This allows selinux support to be disabled even when libselinux is
installed.

Ref: https://wiki.gentoo.org/wiki/Project:Quality_Assurance/Automagic_dependencies

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -22,12 +22,16 @@ dep_csundry = dependency('libcsundry', version: '>=1', fallback: [ 'c-sundry', '
 dep_dbus = dependency('dbus-1', version: '>=1.10', required: false)
 dep_glib = dependency('glib-2.0', version: '>=2.50', required: false)
 dep_libaudit = dependency('audit', version: '>=2.7.0', required: false)
-dep_libselinux = dependency('libselinux', version: '>=2.5', required: false)
 dep_libsystemd = dependency('libsystemd', version: '>=230', required: false)
 dep_math = cc.find_library('m')
 dep_systemd = dependency('systemd', required: false)
 dep_thread = dependency('threads')
 dep_expat = dependency('expat')
+
+use_selinux = get_option('selinux')
+if use_selinux
+        dep_libselinux = dependency('libselinux', version: '>=2.5')
+endif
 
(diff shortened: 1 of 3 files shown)


meson_options.txt
```diff
@@ -0,0 +1,2 @@
+option('selinux', type : 'combo', choices : ['auto', 'true', 'false'],
```
bil-elmoussaoui: Using `value: 'true` to be a default value would be better than having 3 choices

floppym: This is copied straight from systemd. The had reasons for doing it this way.

floppym: Defaulting it to true would cause meson to bail out with an error if libselinux is not installed and the user does not pass false explicitly.

bil-elmoussaoui: Ah i see, thanks for the explanation!

dvdhrm: No auto-detection please. This should be:

`option('selinux', type: 'boolean', value: false)`

We neither use whitespace in front of colons.

meson.build
```diff
@@ -22,13 +22,21 @@ dep_csundry = dependency('libcsundry', version: '>=1', fallback: [ 'c-sundry', '
...
 
+want_selinux = get_option('selinux')
+if want_selinux != 'false'
+        dep_libselinux = dependency('libselinux', version: '>=2.5',
+                                    required: want_selinux == 'true')
+        have_libselinux = dep_libselinux.found()
+else
+        have_libselinux = false
+endif
+
```
dvdhrm: With 'selinux'-option as a boolean, this should be:

```
use_selinux = get_option('selinux')
if use_selinux
        dep_libselinux = dependency('libselinux', version: '>=2.5')
endif
```

src/meson.build
```diff
@@ -49,7 +49,7 @@ else
         ]
 endif
 
-if dep_libselinux.found()
+if have_libselinux
```
dvdhrm: `use_selinux`

floppym:
Thanks for the the feedback. I have force-pushed a new version.

teg:
Works perfectly!

Issue #16, opened 2017-08-16
this might be a good opportunity to replace calls to avc_has_perm() with calls to selinux_access_check() in selinux.c


doverride:
avc_has_perm() is a legacy compatibility interface provided with libselinux, and it should ideally be replaced with calls to selinux_access_check()

as more important components like systemd become more dependent on dbus it becomes more important that the selinux code is up-to-date.

i understand that you may not see it as your duty to address this issue, but i am taking the opportunity to inform you that selinux.c is calling legacy compatibility interfaces, just in case you feel up to the task.

this might be a good opportunity to revisit dbus calls to legacy and compatibility interfaces provided by libselinux

https://github.com/SELinuxProject/selinux/issues/34

dvdhrm:
Thanks for the hint!

`selinux_access_check()` resolves the contexts on every call. `avc_has_perm()` allows us to cache the lookups by storing the sid pointers rather than the context-strings. Is this intentional?

doverride:
On Wed, Aug 16, 2017 at 04:07:14PM +0000, David Herrmann wrote:
> Thanks for the hint!
> 
> `selinux_access_check()` resolves the contexts on every call. `avc_has_perm()` allows us to cache the lookups by storing the sid pointers rather than the context-strings. Is this intentional?

I am not very familiar with all the reasons why selinux_access_check() should be used over avc_has_perm() and if the above has anything to do with the following:

the avc_has_perm() causes "user space object managers" (like dbus) to become confused when there is interference with the ordering of "access vectors", and this might actually be due to that caching aspect (although i am not sure)

what that means in practical terms is that with `avc_has_perm()` one can, at runtime, make dbus confused and block all access if one "interferes with the ordering of access vectors" by for example adding a new access vector at runtime.

this, today, is becoming a greater issue since systemd for example to a greater extend relies on dbus.

so if i interfere with the ordering of access vectors at runtime by adding a new policy module that declares a new access vector for example, then dbus potentially becomes confused (possibly due to the caching aspect you described) and blocks all access. that means that systemd can no longer do its job properly and well, if systemd can't do its job then theres a problem.

By the way, systemd also uses `selinux_access_check()`. So you may find examples of its usage there

> 
> -- 
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/bus1/dbus-broker/issues/16#issuecomment-322820695

-- 
[...]

PR #34, opened 2017-08-25
util/selinux: move from using security IDs to using security contexts

teg:
The SID API is deprecated upstream, instead move to the
selinux_check_access() API which does the lookup from context to SID
internally. If this proves to have a significant overhead we should
work with upstream SELinux to come up with a non-deprecated solution.

Addresses issue #16.

### DIFF
--- a/src/broker/controller.h
+++ b/src/broker/controller.h
@@ -15,7 +15,6 @@
 
 typedef struct Broker Broker;
 typedef struct Bus Bus;
-typedef struct BusSELinuxID BusSELinuxID;
 typedef struct Controller Controller;
 typedef struct ControllerName ControllerName;
 typedef struct ControllerListener ControllerListener;
@@ -63,7 +62,8 @@ struct ControllerListener {
 
 struct Controller {
         Broker *broker;
-        BusSELinuxID *sid;
+        char *seclabel;
+        size_t n_seclabel;
         Connection connection;
         CRBTree name_tree;
         CRBTree listener_tree;
--- a/src/bus/policy.c
+++ b/src/bus/policy.c
@@ -302,7 +302,7 @@ static int policy_registry_node_new(PolicyRegistryNode **nodep, CRBTree *tree, u
 /**
  * policy_registry_new() - XXX
  */
-int policy_registry_new(PolicyRegistry **registryp, BusSELinuxID *fallback_id) {
+int policy_registry_new(PolicyRegistry **registryp, const char *fallback_seclabel) {
         _c_cleanup_(policy_registry_freep) PolicyRegistry *registry = NULL;
         int r;
 
@@ -312,7 +312,7 @@ int policy_registry_new(PolicyRegistry **registryp, BusSELinuxID *fallback_id) {
 
         *registry = (PolicyRegistry)POLICY_REGISTRY_NULL;
 
-        r = bus_selinux_registry_new(&registry->selinux, fallback_id);
+        r = bus_selinux_registry_new(&registry->selinux, fallback_seclabel);
         if (r)
                 return error_fold(r);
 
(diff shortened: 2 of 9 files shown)


src/bus/policy.c
```diff
@@ -563,7 +564,10 @@ int policy_snapshot_new(PolicySnapshot **snapshotp,
         *snapshot = (PolicySnapshot)POLICY_SNAPSHOT_NULL;
 
         snapshot->selinux = bus_selinux_registry_ref(registry->selinux);
-        snapshot->sid = sid;
+
+        context_new = strdup(context);
+        if (!context)
```
Keruspe: Shouldn't this be !context_new ?

teg: Ouch! Thanks!

src/broker/controller.h
```diff
@@ -63,7 +62,7 @@ struct ControllerListener {
 
 struct Controller {
         Broker *broker;
-        BusSELinuxID *sid;
+        char *context;
```
dvdhrm: `context`? Can't we come up with something more descriptive? I'd really prefer secctx or seclabel. You already call it `seclabel` on the peer, so why not here?

Also, are you sure we can treat *all* seclabels as 0-terminated strings? So far we always tried to treat them as sized blob and just added a 0-termination to make them pseudo-printable. I am aware that most (all?) LSMs use strings there, but is that guaranteed?

src/bus/policy.c
```diff
@@ -302,7 +302,7 @@ static int policy_registry_node_new(PolicyRegistryNode **nodep, CRBTree *tree, u
 /**
  * policy_registry_new() - XXX
  */
-int policy_registry_new(PolicyRegistry **registryp, BusSELinuxID *fallback_id) {
+int policy_registry_new(PolicyRegistry **registryp, const char *fallback_context) {
```
dvdhrm: `fallback_seclabel`?

src/bus/policy.c
```diff
@@ -563,7 +564,11 @@ int policy_snapshot_new(PolicySnapshot **snapshotp,
         *snapshot = (PolicySnapshot)POLICY_SNAPSHOT_NULL;
 
         snapshot->selinux = bus_selinux_registry_ref(registry->selinux);
-        snapshot->sid = sid;
+
+        context_new = strdup(context);
+        if (!context_new)
+                return error_origin(-ENOMEM);
+        snapshot->context = context_new;
```
dvdhrm: Why the temporary variable? Why not this:

```c
snapshot->seclabel = strdup(seclabel);
if (!snapshot->seclabel)
        return error_origin(-ENOMEM);
```

src/bus/policy.c
```diff
@@ -613,7 +620,12 @@ int policy_snapshot_dup(PolicySnapshot *snapshot, PolicySnapshot **newp) {
         *new = (PolicySnapshot)POLICY_SNAPSHOT_NULL;
 
         new->selinux = bus_selinux_registry_ref(snapshot->selinux);
-        new->sid = snapshot->sid;
+
+        context_new = strdup(snapshot->context);
+        if (!context_new)
+                return error_origin(-ENOMEM);
+        new->context = context_new;
+        context_new = NULL;
```
dvdhrm: Same as above: Why the temporary variable?

src/util/selinux.c
```diff
@@ -109,19 +61,20 @@ C_DEFINE_CLEANUP(BusSELinuxName *, bus_selinux_name_free);
...
+        selinux_name->context = NULL;
         memcpy(selinux_name->name, name, n_name);
 
-        r = avc_context_to_sid(context, &selinux_name->sid);
-        if (r < 0)
-                return error_origin(-errno);
+        context_new = strdup(context);
+        if (!context_new)
+                return error_origin(-ENOMEM);
+        selinux_name->context = context_new;
```
dvdhrm: Same: Why not assign it directly?

src/broker/controller.h
teg: I'll rename to seclabel. I could then also store n_seclabel, though it is totally fine to rely on these things being null-terminated as our helpers guarantee this (and in particular libselinux requires this). Othe LMSs could in principle have seclablels that contain null-bytes, in which case the length would be needed if we ever made use of it, so I'll add it now for completeness.

Issue #28, opened 2017-08-24
activation of a masked unit results in long timeout


michich:
Description:
After I switched to dbus-broker, logging in to Gnome Shell takes a long time (tens of seconds), and so does starting any application. I traced the problem to activation of masked units. Instead of failing immediately, such activation requests time out.

Steps to reproduce:
systemctl --user mask --now at-spi-dbus-bus.service
time busctl --user get-property org.a11y.Bus /org/a11y/bus org.a11y.Status ScreenReaderEnabled

Actual result:
Connection timed out
real    0m25.032s

Expected result (actual result when using dbus-daemon):
Unit at-spi-dbus-bus.service is masked.
real    0m0.005s

Additional information:
dbus-broker-launch sends the activation request signal ("org.freedesktop.systemd1.Activator", "ActivationRequest") to systemd, but it implements no handling of systemd's signal about the activation failure ("org.freedesktop.systemd1.Activator", "ActivationFailure").

teg:
Thanks for the report, and your analysis is correct, the handling of ActivationFailure is indeed missing.

teg:
This should now be fixed in git. Closing.

PR #42, opened 2017-08-28
Properly fail activation requests if systemd cannot activate them

teg:
This can happen if the unit in question has been masked (or if it does not exist, but that would be a configuration failure). In this case, the requesting client should not hang forever, but rather receive a proper error reply.

This addresses issue #28.

### DIFF
--- a/src/broker/controller.c
+++ b/src/broker/controller.c
@@ -9,6 +9,7 @@
 #include "broker/controller.h"
 #include "bus/activation.h"
 #include "bus/bus.h"
+#include "bus/driver.h"
 #include "bus/policy.h"
 #include "dbus/connection.h"
 #include "dbus/message.h"
@@ -64,8 +65,14 @@ static int controller_name_new(ControllerName **namep, Controller *controller, c
 /**
  * controller_name_reset() - XXX
  */
-void controller_name_reset(ControllerName *name) {
-        activation_flush(&name->activation);
+int controller_name_reset(ControllerName *name) {
+        int r;
+
+        r = driver_name_activation_failed(&name->controller->broker->bus, &name->activation);
+        if (r)
+                return error_fold(r);
+
+        return 0;
 }
 
 /**
--- a/src/launch/main.c
+++ b/src/launch/main.c
@@ -33,6 +33,7 @@ enum {
 
 struct Service {
         Manager *manager;
+        sd_bus_slot *slot;
         CRBNode rb;
         char *name;
         char *unit;
@@ -94,6 +95,7 @@ static Service *service_free(Service *service) {
         free(service->exec);
         free(service->unit);
         free(service->name);
+        sd_bus_slot_unref(service->slot);
         free(service);
 
         return NULL;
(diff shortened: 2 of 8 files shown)


src/broker/controller.c
```diff
@@ -64,8 +65,14 @@ static int controller_name_new(ControllerName **namep, Controller *controller, c
 /**
  * controller_name_reset() - XXX
  */
-void controller_name_reset(ControllerName *name) {
-        activation_flush(&name->activation);
```
dvdhrm: Can you make `activation_flush()` static now? Not used by anyone but activation.c.

src/launch/main.c
```diff
@@ -375,6 +375,8 @@ static int manager_start_transient_unit(Manager *manager, Service *service) {
                                            "/org/freedesktop/systemd1",
                                            "org.freedesktop.systemd1.Manager",
                                            "StartTransientUnit");
+        if (r < 0)
+                return error_origin(r);
```
dvdhrm: Oops!

src/launch/main.c
```diff
@@ -332,26 +332,26 @@ static int manager_fork(Manager *manager, int fd_controller) {
...
+        r = sd_bus_message_new_method_call(manager->bus_regular, &method_call,
+                                           "org.freedesktop.systemd1",
+                                           "/org/freedesktop/systemd1",
+                                           "org.freedesktop.systemd1.Manager",
+                                           "StartUnit");
         if (r < 0)
                 return error_origin(r);
 
-        r = sd_bus_message_set_destination(signal, "org.freedesktop.systemd1");
+        r = sd_bus_message_append(method_call, "ss", service->unit, "fail");
```
dvdhrm: This needs to be "replace".

src/launch/main.c
```diff
@@ -464,7 +503,7 @@ static int manager_start_transient_unit(Manager *manager, Service *service) {
         if (r < 0)
                 return error_origin(r);
 
-        r = sd_bus_send(manager->bus_regular, method_call, NULL);
+        r = sd_bus_call_async(manager->bus_regular, &service->slot, method_call, manager_start_unit_handler, service, -1);
```
dvdhrm: Ehhh, in both cases, you sure this is safe? I mean, what guarantees you that `service->slot` is NULL and can be re-used?

Imagine this: Name is activated. We get the signal and send the request to systemd. In The meantime, the name is dropped and activated again. We get the signal again and will reuse `service->slot` before it was cleared by the reply handler.

I think the right solution is to do `service->slot = sd_bus_slot_unref(service->slot);` right at the top of `manager_start_unit()`. So we discard the old request and send a new-one. We cannot just skip it, since someone might rely on a barrier and changed systemd units etc. So we must send a new request to guarantee ordering.

src/launch/main.c
```diff
@@ -585,7 +599,7 @@ static int manager_on_set_activation_environment(Manager *manager, sd_bus_messag
         if (r < 0)
                 return error_origin(r);
 
-        r = sd_bus_send(manager->bus_regular, method_call, NULL);
+        r = sd_bus_call_async(manager->bus_regular, NULL, method_call, manager_set_environment_handler, NULL, -1);
```
dvdhrm: Urgs, what are the lifetime rules for floating slots again? They're destroyed when we drop our last bus-reference? If that's true, then this looks ok. I know, we don't pass a userdata context, but still...

src/broker/controller.c
teg: Done.

src/launch/main.c
teg: I don't see the issue either way. There is nothing for the slot to protect... But yes, lifetime is the same as the bus, or the timeout (which we set to -1).

src/launch/main.c
teg: Broker will not send more than one request at a time, but you are right, the service can be activated, grab its name, release it again, and someone can activate the name again, before we get the answer to StartUnit() from systemd, and this would leak the slot. I'll do clear it as you suggested.

PR #46, opened 2017-09-01
Log Infrastructure

dvdhrm:
This adds the discussed log infrastructure, allowing us to push messages to systemd-journal or stderr, depending on which one is configured.

See the first patch for details on the infrastructure itself. The other patches just add the integration to dbus-broker and dbus-broker-launch. The idea there is to open the log FD in the launcher and pass it to dbus-broker just like the controller FD. By default, logging is disabled, so our test infrastructure still works as expected.

So far, we only initialize the log-context on `struct Broker`, but no messages are printed. This needs to be done as a follow-up.

### DIFF
--- a/src/util/log.c
+++ b/src/util/log.c
@@ -1,0 +1,660 @@
+/*
+ * Log Context
+ *
+ * The log context provides an infrastructure to send log-messages to the
+ * system log daemon. Both structured and unstructured logging is supported.
+ * Depending on which logging daemon is used, a compatible mode is selected.
+ *
+ * Right now, these modes are supported:
+ *
+ *     * stderr: Stream-based, unstructured logging to the inherited stderr
+ *               channel.
+ *               This is the default.
+ *
+ *     * journal: Datagram-based, structured logging to the journal. Other
+ *                journal-API-compatible daemons can be used as well. No
+ *                journal-specific API besides the datagram socket is used.
+ *
+ * The logging API provides a staging buffer to assemble structured log
+ * messages. Once a message is complete, it must be committed, which will send
+ * it out and prepare the context for the next message.
+ *
+ * The log_*append*() APIs append log fields to the staging buffer. They can
+ * be used consecutively. Convenience wrappers are provided to add the most
+ * common structured fields. The staging buffer should only be used for
+ * structured logging. The main log-message is not part of it.
+ *
+ * Once a log message is completed, the log_*commit*() APIs commit the log
+ * message. At the time of commit, the caller must provide the actual main log
+ * message, which appears in the logs. Multiple convenience wrappers are
+ * provided to allow formatted messages.
+ * Note that depending on the backend, the structured fields might be
+ * discarded. Not all logging systems support them.
+ *
+ * By default, logging is synchronous and blocking. That is, no data is
+ * buffered in the log-context once a message is committed. All data is written
+ * into the socket in a blocking manner. The data will be buffered in the
+ * kernel socket queues, of course.
+ * This mode is not always desirable. In particular, we allow logging on behalf
+ * of peers in dbus-broker. This means, whenever a peer screws up, we often
+ * have more information than the peer itself, so we want to log data on behalf
+ * of them. This, however, makes us susceptible to DoS attacks. Hence, we
+ * support a lossy mode. If a log context is set to lossy mode, messages will
+ * be submitted to the socket in non-blocking mode. Once the kernel buffers run
+ * full, we submit a single, one-time, synchronous warning to the log, and from
+ * then on will discard log messages, if the kernel buffers are full.
+ *
+ * Lastly, please note that each log context must not be used from multiple
+ * threads. Use separate contexts for each thread. Also be aware that stream
+ * sockets do not have atomic writes, so you might need separate sockets.
+ *
+ * In case of structured logging, we support:
+ *
+ *     * MESSAGE: The string-formatted log message.
+ *
+ *     * CODE_FILE: Source-code file where the logging call originated. Usually
+ *                  corresponds to the expanded __FILE__ constant.
+ *
+ *     * CODE_LINE: Source-code line number where the logging call originated.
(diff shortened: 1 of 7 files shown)


src/util/log.c
```diff
@@ -0,0 +1,648 @@
...
+         * This guarantees the operation will be non-blocking, as long as
+         * suitably sized log messages are used.
+         */
+
+        if (log->lossy) {
+                if (log->n_dropped) {
+                        r = ioctl(log->log_fd, SIOCOUTQ, &v);
+                        if (r < 0) {
+                                return error_origin(-errno);
+                        } else if (v) {
```
teg: Shouldn't we be checking not for the queue being empty, but if the message will fit. Otherwise you have 1) dropping messages unnecessarily and 2) writing out partial messages that are too big to fit even in an empty queue (and hence blocking later in the function).

src/util/log.c
```diff
@@ -0,0 +1,648 @@
...
+                }
+
+                l = send(log->log_fd, blob, n_blob, MSG_NOSIGNAL | MSG_DONTWAIT);
+                if (l >= (ssize_t)n_blob) {
+                        return 0;
+                } else if (l < 0) {
+                        if (errno != EAGAIN)
+                                return error_origin(-errno);
+
+                        l = 0;
```
teg: So in case of a short write, we resubmt the prefix that was already pushed to the socket?

src/util/log.c
```diff
@@ -0,0 +1,648 @@
...
+         * We could not send the message as a single datagram. Instead, we now
+         * seal the memfd and send it as payload of an empty datagram. The
+         * journal can deal with this and treats the memfd content as log
+         * message.
+         *
+         * Note that this means there is an inflight memfd pending on the
+         * journal that is accounted on us as long as the journal did not read
+         * it, yet. Hence, clients should never be able to trigger such log
+         * messages, otherwise they could exploit this and make us exceed our
+         * inflight FD limit.
```
teg: Shouldn't this be enforced by having to enable some special mode? Otherwise, how is the user of the logging API supposed to know exactly how much stuff they can put in a message? Could even make this conditional on `!lossy`?

src/util/log.c
```diff
@@ -0,0 +1,648 @@
...
+        log->offset = 0;
+
+        if (format && *format) {
+                log_appendf(log, "<%d>", log->level ?: 6);
+                log_vappendf(log, format, args);
+                log_appends(log, "\n");
+        }
+
+        if (log->error)
+                r = error_trace(log->error);
```
teg: Should the poison be reset at this point?

src/util/log.c
```diff
@@ -0,0 +1,648 @@
...
+
+        switch (log->mode) {
+        case LOG_MODE_STDERR:
+                r = log_commit_stderr(log, format, args);
+                break;
+        case LOG_MODE_JOURNAL:
+                r = log_commit_journal(log, format, args);
+                break;
+        default:
+                r = error_trace(log->error);
```
teg: Why this error? And in case `log->error` is 0, should we not be returning an error regardless in this case?

src/util/log.c
```diff
@@ -0,0 +1,648 @@
...
+                break;
+        }
+
+        log->error = 0;
+        log->level = 0;
+        log->offset = 0;
```
teg: See above, shouldn't these things be reinitialized in the specific handlers?

src/util/log.c
```diff
@@ -0,0 +1,648 @@
...
+         * ignore anything that is in the staging buffer and instead just print
+         * our message.
+         * Sadly, `%r' is still no standardized format specifier, so we need
+         * the separate calls into log_append() here.
+         */
+
+        log->offset = 0;
+
+        if (format && *format) {
+                log_appendf(log, "<%d>", log->level ?: 6);
```
teg: Why the magic number? Make a define? Is this not LOG_INFO? (same elsewhere). This is basically the same logic as in log_append_common(), right?

src/util/log.c
dvdhrm: How?

src/util/log.c
dvdhrm: Errr, what prefix was already pushed?

src/util/log.c
dvdhrm: The default is /dev/null, which is handled here. Why should we return an error?

src/util/log.c
dvdhrm: Why?

Whenever you call log_commit(), the current log message is dealt with and the log context is prepared for the next log message, regardless whether it fails or not. Nobody will "try" to commit a message and then want to continue working on that message. A commit is a one-shot operation that either submits or discards the message. Hence, I reset this in log_commit() to make sure all paths deal with this correctly.

src/util/log.c
dvdhrm: Conditional on `lossy` sounds fine to me.

src/util/log.c
teg: Make it an explicit case? Just to make it more obvious that this is expected.

src/util/log.c
teg: Fair enough.

src/util/log.c
teg: Ignore that, misread.

src/util/log.c
teg: Just add a comment to explain why it is not possible (if you agree that that's what we would have liked to do). Also, isn't 2) a problem? What's stopping someone from repeatedly triggering overly long log messages that will end up blocking every time the output queue is empty?

dvdhrm: How can you trigger overly long log messages? This is the stderr-backend, it uses just the MESSAGE= part. That should always fit into the kernel buffer. If not, you're doing something wrong, aren't you?

teg: Ah, missed that part. That's fine then. Comment would be nice.

PR #51, opened 2017-09-04
Add basic logging infrastructure

teg:
We should log if any message deliver fails unexpectedly. For now, add the logging for policy failures, which is most likely the most fragile and hard to debug. In the future we will add more metadata to the logging and extend it to log in more cases.

### DIFF
--- a/src/broker/broker.c
+++ b/src/broker/broker.c
@@ -83,7 +83,7 @@ int broker_new(Broker **brokerp, int log_fd, int controller_fd, uint64_t max_byt
         else
                 return error_origin(-ENOTRECOVERABLE);
 
-        r = bus_init(&broker->bus, max_bytes, max_fds, max_matches, max_objects);
+        r = bus_init(&broker->bus, &broker->log, max_bytes, max_fds, max_matches, max_objects);
         if (r)
                 return error_fold(r);
 
--- a/src/broker/broker.c
+++ b/src/broker/broker.c
@@ -83,6 +83,9 @@ int broker_new(Broker **brokerp, int log_fd, int controller_fd, uint64_t max_byt
         else
                 return error_origin(-ENOTRECOVERABLE);
 
+        /* XXX: make this run-time optional */
+        log_set_lossy(&broker->log, true);
+
         r = bus_init(&broker->bus, &broker->log, max_bytes, max_fds, max_matches, max_objects);
         if (r)
                 return error_fold(r);
--- a/src/bus/bus.c
+++ b/src/bus/bus.c
@@ -12,6 +12,7 @@
 #include "bus/name.h"
 #include "dbus/address.h"
 #include "util/error.h"
+#include "util/log.h"
 #include "util/user.h"
 
 int bus_init(Bus *bus,
@@ -76,4 +77,45 @@ Peer *bus_find_peer_by_name(Bus *bus, Name **namep, const char *name_str) {
         if (namep)
                 *namep = name;
         return peer;
+}
+
+static int bus_log_commit_policy(Bus *bus, const char *action, uint64_t sender_id, uint64_t receiver_id, Message *message) {
+        Log *log = bus->log;
+        int r;
+
+        message_log_append(message, log);
+
+        log_appendf(log, "DBUS_BROKER_TRANSMIT_ACTION=%s\n", action);
+
+        if (sender_id == ADDRESS_ID_INVALID) {
+                log_appendf(log,
+                            "DBUS_BROKER_SENDER_UNIQUE_NAME=org.freedesktop.DBus\n");
+        } else {
+                log_appendf(log,
+                            "DBUS_BROKER_SENDER_UNIQUE_NAME=:1.%llu\n",
+                            sender_id);
+        }
+
+        if (receiver_id == ADDRESS_ID_INVALID) {
+                log_appendf(log,
+                            "DBUS_BOKER_RECEIVER_UNIQUE_NAME=org.freedesktop.DBus\n");
+        } else {
+                log_appendf(log,
+                            "DBUS_BROKER_RECEIVER_UNIQUE_NAME=:1.%llu\n",
+                            receiver_id);
+        }
+
+        r = log_commitf(log, ":1.%llu failed to %s message, due to policy.", sender_id, action);
+        if (r)
+                return error_fold(r);
+
+        return 0;
+}
+
+int bus_log_commit_policy_send(Bus *bus, uint64_t sender_id, uint64_t receiver_id, Message *message) {
+        return bus_log_commit_policy(bus, "send", sender_id, receiver_id, message);
+}
+
+int bus_log_commit_policy_receive(Bus *bus, uint64_t receiver_id, uint64_t sender_id, Message *message) {
+        return bus_log_commit_policy(bus, "receive", sender_id, receiver_id, message);
 }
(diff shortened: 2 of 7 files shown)


src/broker/broker.c
```diff
@@ -92,6 +92,8 @@ int broker_new(Broker **brokerp, int log_fd, int controller_fd, uint64_t max_byt
         if (r)
                 return error_fold(r);
 
+        broker->bus.log = &broker->log;
```
dvdhrm: Now there is a random time between bus_init() and this where bus->log is NULL. This will not work as expected. Why not pass it in to bus_init()?

src/dbus/message.c
```diff
@@ -598,3 +599,52 @@ void message_stitch_sender(Message *message, uint64_t sender_id) {
...
+                            "MESSAGE_REPLY_SERIAL=%d\n",
+                            message->metadata.fields.reply_serial);
+                break;
+        case DBUS_MESSAGE_TYPE_ERROR:
+                log_appendf(log,
+                            "MESSAGE_TYPE=method_return\n"
+                            "MESSAGE_ERROR_NAME=%s\n"
+                            "MESSAGE_REPLY_SERIAL=%d\n",
+                            message->metadata.fields.error_name,
+                            message->metadata.fields.reply_serial);
```
dvdhrm: I'd prefer not picking random fields in a global context. So far I used `DBUS_BROKER_*` prefix for fields we "invent". Furthermore, can we document them (at least mention them) in `src/util/log.c`?

src/bus/bus.h
```diff
@@ -60,3 +60,12 @@ int bus_init(Bus *bus,
...
 Peer *bus_find_peer_by_name(Bus *bus, Name **namep, const char *name);
+
+int bus_log_policy_send(Bus *bus, uint64_t sender_id, uint64_t receiver_id, Message *message, const char *file, int line, const char *func);
+int bus_log_policy_receive(Bus *bus, uint64_t sender_id, uint64_t receiver_id, Message *message, const char *file, int line, const char *func);
+
+#define bus_log_policy_send_here(_bus, _sender_id, _receiver_id, _message) \
+        bus_log_policy_send((_bus), (_sender_id), (_receiver_id), (_message), __FILE__, __LINE__, __func__)
+
+#define bus_log_policy_receive_here(_bus, _receiver_id, _sender_id, _message) \
+        bus_log_policy_receive((_bus), (_receiver_id), (_sender_id), (_message), __FILE__, __LINE__, __func__)
```
dvdhrm: Ugh. Can't we call log_append_here() in the caller? And then, for consistency, call these `bus_log_commit_policy_send()`? Basically following the style of `foo_log_{append,commit}_bar()`.

src/bus/bus.c
```diff
@@ -75,3 +76,65 @@ Peer *bus_find_peer_by_name(Bus *bus, Name **namep, const char *name_str) {
...
+                            "SENDER_UNIQUE_NAME=:1.%llu\n",
+                            sender_id);
+        }
+
+        r = log_commitf(log, "Message could not be received by :1.%llu, due to policy.", receiver_id);
+        if (r)
+                return error_fold(r);
+
+        return 0;
+}
```
dvdhrm: Can we make this one single function? This seems to duplicate quite a bit of code.

src/bus/peer.c
```diff
@@ -605,7 +615,7 @@ int peer_queue_reply(Peer *sender, const char *destination, uint32_t reply_seria
         return 0;
 }
 
-static int peer_broadcast_to_matches(PolicySnapshot *sender_policy, NameSet *sender_names, MatchRegistry *matches, MatchFilter *filter, uint64_t transaction_id, Message *message) {
+static int peer_broadcast_to_matches(PolicySnapshot *sender_policy, NameSet *sender_names, MatchRegistry *matches, uint64_t sender_id, MatchFilter *filter, uint64_t transaction_id, Message *message) {
```
dvdhrm: Now the argument order looks pretty random, or is there some logic behind it? I'd do something like `sender-id, sender-creds..., matches, message`

src/dbus/message.c
teg: I added DBUS_BROKER_* prefix to everything now. However, is it not odd to document this in util/? Maybe it should be documented, but shouldn't util be generic? And wouldn't perhaps the manpage be the better place?

src/bus/bus.c
teg: Done.

src/bus/peer.c
teg: Just followed peer_broadcast(), but changed now.

src/bus/bus.h
teg: Done.

src/broker/broker.c
teg: Done.

src/dbus/message.c
dvdhrm: Yeah, maybe we should rename `DBUS_BROKER_LOG_DROPPED` to something official, so util/log.c is dbus-broker-independent, and then move the docs to the wiki?

src/dbus/message.c
```diff
@@ -11,6 +11,7 @@
 #include "dbus/message.h"
 #include "dbus/protocol.h"
 #include "util/fdlist.h"
+#include "util/log.h"
 #include "util/error.h"
```
dvdhrm: My eyes! They hurt! This is not sorted alphanumerically... (I will push a fix myself)

src/dbus/message.c
```diff
@@ -598,3 +599,52 @@ void message_stitch_sender(Message *message, uint64_t sender_id) {
         else
                 message->header->n_fields = htole32(message->n_header - sizeof(*message->header));
 }
+
+void message_log_append(Message *message, Log *log) {
+        log_appendf(log,
+                    "DBUS_BROKER_MESSAGE_DESTINATION=%s\n"
+                    "DBUS_BRKOER_MESSAGE_SERIAL=%d\n"
```
dvdhrm: `BRKOER`?

I will push a fix myself.

src/dbus/message.c
```diff
@@ -598,3 +599,52 @@ void message_stitch_sender(Message *message, uint64_t sender_id) {
...
+                    "DBUS_BROKER_MESSAGE_SIGNATURE=%s\n",
+                    message->metadata.fields.destination ?: "<broadcast>",
+                    message->metadata.header.serial,
```
dvdhrm: This is `uint32_t` isn't it? I will fix the format-string.

src/dbus/message.c
```diff
@@ -598,3 +599,52 @@ void message_stitch_sender(Message *message, uint64_t sender_id) {
...
+                    message->metadata.fields.signature ?: "<missing>");
+
+        switch (message->metadata.header.type) {
+        case DBUS_MESSAGE_TYPE_METHOD_CALL:
+                log_appendf(log,
+                            "DBUS_BROKER_MESSAGE_TYPE=method_call\n"
+                            "DBUS_BROKER_MESSAGE_PATH=%s\n"
+                            "DBUS_BRKOER_MESSAGE_INTERFACE=%s\n"
```
dvdhrm: `BRKOER`?

src/dbus/message.c
dvdhrm: Btw., what is wrong with just appending all **known** header-fields unconditionally?

Issue #48, opened 2017-09-02
[RFE] [build] generate/install manpage from meson


ignatenkobrain:
It is much nicer way of doing thing than doing in spec file (think about other distributions than Fedora/*EL).

dvdhrm:
Done.

Please re-open if anything still needs to be fixed!

PR #60, opened 2017-09-19
build: generate manpages

teg:
This was done manually in the specfile, let meson do it.

This addresses issue #48.

### DIFF
--- a/docs/meson.build
+++ b/docs/meson.build
@@ -1,0 +1,19 @@
+custom_target(
+        'man_dbus_broker',
+        output: 'dbus-broker.1',
+        input: 'dbus-broker.rst',
+        command: [rst2man, '@INPUT@', '@OUTPUT@'],
+        install: true,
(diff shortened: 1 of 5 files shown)


docs/meson.build
```diff
@@ -0,0 +1,19 @@
+custom_target(
+        'man_dbus_broker',
+        output : 'dbus-broker.1',
+        input : 'dbus-broker.rst',
+        command : [rst2man, '@INPUT@', '@OUTPUT@'],
+        install : true,
+        install_dir : join_paths(get_option('mandir'), 'man1')
```
dvdhrm: We don't put spaces before the colon ;)

Also, doesn't this need: `build_by_default: true`
Or is the `install: true` enough to pull this in?

teg: It appears to be enough.

teg: Fixed the whitespace.

teg:
Added python-docutils to build requirements (which provides rst2man).

PR #62, opened 2017-09-21
uid ranges

teg:
Allow pseudo-compatibility with at_console= policies, by considering all system users at_console=false and all regular users at_console=true. This is done by allowing the UID specified in the broker API to be an upper/lower bound in addition to an exact match.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -42,6 +42,7 @@ if use_launcher
         dep_libsystemd = dependency('libsystemd', version: '>=230')
         dep_systemd = dependency('systemd', version: '>=230')
 
+        add_project_arguments('-DSYSTEMUIDMAX=' + dep_systemd.get_pkgconfig_variable('systemuidmax'), language: 'c')
         conf.set('systemunitdir', dep_systemd.get_pkgconfig_variable('systemdsystemunitdir'))
         conf.set('userunitdir', dep_systemd.get_pkgconfig_variable('systemduserunitdir'))
 endif
--- a/src/bus/policy.c
+++ b/src/bus/policy.c
@@ -481,7 +481,6 @@ static int policy_registry_import_batch(PolicyRegistry *registry,
  */
 int policy_registry_import(PolicyRegistry *registry, CDVar *v) {
         PolicyRegistryNode *node;
-        uint32_t uidgid;
         int r;
 
         /* XXX: provide the type */
@@ -494,9 +493,15 @@ int policy_registry_import(PolicyRegistry *registry, CDVar *v) {
         c_dvar_read(v, "[");
 
         while (c_dvar_more(v)) {
-                c_dvar_read(v, "(u", &uidgid);
+                uint32_t uid_start, uid_end;
 
-                r = policy_registry_at_uid(registry, &node, uidgid);
+                c_dvar_read(v, "(uu", &uid_start, &uid_end);
+
+                /* XXX: support ranges */
+                if (uid_start != uid_end)
+                        return POLICY_E_INVALID;
+
+                r = policy_registry_at_uid(registry, &node, uid_start);
                 if (r)
                         return error_trace(r);
 
@@ -510,9 +515,11 @@ int policy_registry_import(PolicyRegistry *registry, CDVar *v) {
         c_dvar_read(v, "][");
 
         while (c_dvar_more(v)) {
-                c_dvar_read(v, "(u", &uidgid);
+                uint32_t gid;
 
-                r = policy_registry_at_gid(registry, &node, uidgid);
+                c_dvar_read(v, "(u", &gid);
+
+                r = policy_registry_at_gid(registry, &node, gid);
                 if (r)
                         return error_trace(r);
 
--- a/src/bus/policy.c
+++ b/src/bus/policy.c
@@ -255,11 +255,15 @@ static int policy_batch_add_recv(PolicyBatch *batch,
 
 static int policy_registry_node_compare(CRBTree *t, void *k, CRBNode *n) {
         PolicyRegistryNode *node = c_container_of(n, PolicyRegistryNode, registry_node);
-        uint32_t uidgid = (uint32_t)(unsigned long)k;
+        PolicyRegistryNodeIndex *index = k;
 
-        if (uidgid < node->uidgid)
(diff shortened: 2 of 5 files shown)


src/launch/policy.c
```diff
@@ -211,12 +220,12 @@ static int policy_at_uidgid(CRBTree *tree, PolicyNode **nodep, uint32_t uidgid)
...
-static int policy_at_uid(Policy *policy, PolicyNode **nodep, uint32_t uidgid) {
-        return policy_at_uidgid(&policy->uid_tree, nodep, uidgid);
+static int policy_at_uid(Policy *policy, PolicyNode **nodep, uint32_t uid_start, uint32_t uid_end) {
+        return policy_at_uidgid(&policy->uid_tree, nodep, uid_start, uid_end);
 }
 
-static int policy_at_gid(Policy *policy, PolicyNode **nodep, uint32_t uidgid) {
-        return policy_at_uidgid(&policy->uid_tree, nodep, uidgid);
+static int policy_at_gid(Policy *policy, PolicyNode **nodep, uint32_t gid) {
+        return policy_at_uidgid(&policy->uid_tree, nodep, gid, gid);
```
dvdhrm: This needs to be `gid_tree`, doesn't it? Was wrong before. We should fix this now and rebase this on top.

src/bus/policy.c
```diff
@@ -386,11 +407,15 @@ static int policy_registry_at_uidgid(CRBTree *tree, PolicyRegistryNode **nodep,
 }
 
 static int policy_registry_at_uid(PolicyRegistry *registry, PolicyRegistryNode **nodep, uint32_t uid) {
-        return policy_registry_at_uidgid(&registry->uid_tree, nodep, uid);
+        return policy_registry_at_uidgid(&registry->uid_tree, nodep, uid, uid);
+}
+
+static int policy_registry_at_uid_range(PolicyRegistry *registry, PolicyRegistryNode **nodep, uint32_t uid_start, uint32_t uid_end) {
+        return policy_registry_at_uidgid(&registry->uid_tree, nodep, uid_start, uid_end);
```
dvdhrm: This needs to be `uid_range_tree`.

meson.build
```diff
@@ -42,6 +42,7 @@ if use_launcher
         dep_libsystemd = dependency('libsystemd', version: '>=230')
         dep_systemd = dependency('systemd', version: '>=230')
 
+        add_project_arguments('-DSYSTEMUIDMAX="' + dep_systemd.get_pkgconfig_variable('systemuidmax') + '"', language: 'c')
```
dvdhrm: Why a string? Why not:

```
add_project_arguments('-DSYSTEMUIDMAX=' + dep_systemd.get_pkgconfig_variable('systemuidmax'), language: 'c')
```

src/launch/policy.c
```diff
@@ -533,6 +559,18 @@ static int policy_import_recv(Policy *policy, ConfigNode *cnode) {
...
 
                 c_list_link_tail(&node->recv_list, &record->link);
+        } else if (cnode->parent->policy.context == CONFIG_POLICY_NO_CONSOLE) {
+                r = policy_at_uid(policy, &node, 0, strtoul(SYSTEMUIDMAX, NULL, 10));
+                if (r)
+                        return error_trace(r);
+
+                c_list_link_tail(&node->recv_list, &record->link);
+        } else if (cnode->parent->policy.context == CONFIG_POLICY_AT_CONSOLE) {
+                r = policy_at_uid(policy, &node, strtoul(SYSTEMUIDMAX, NULL, 10) + 1, -1);
```
dvdhrm: With `SYSTEMUIDMAX` being an integer, you should be able to drop all these `strtoul` calls.

Also, how about `policy_at_systemuid()` as a small helper that hard-codes 0-SYSTEMUIDMAX?

src/launch/policy.c
```diff
@@ -305,8 +305,10 @@ static int policy_import_connect(Policy *policy, ConfigNode *cnode) {
         }
 
         if (cnode->parent->policy.context == CONFIG_POLICY_USER ||
-            cnode->parent->policy.context == CONFIG_POLICY_GROUP) {
-                fprintf(stderr, "Connection policy not allowed in user/group context in %s +%lu\n",
+            cnode->parent->policy.context == CONFIG_POLICY_GROUP ||
+            cnode->parent->policy.context == CONFIG_POLICY_NO_CONSOLE ||
+            cnode->parent->policy.context == CONFIG_POLICY_AT_CONSOLE) {
```
dvdhrm: I don't think that's correct. Connect-policies are allowed in at_console. Grep for `BUS_POLICY_RULE_IS_PER_CLIENT` in dbus-daemon.

src/bus/policy.c
```diff
@@ -602,20 +612,29 @@ int policy_snapshot_new(PolicySnapshot **snapshotp,
...
-        ++snapshot->n_batches;
+                snapshot->batches[snapshot->n_batches++] = policy_batch_ref(registry->default_batch);
 
         while (n_gids-- > 0) {
                 node = policy_registry_find_gid(registry, gids[n_gids]);
                 if (node)
                         snapshot->batches[snapshot->n_batches++] = policy_batch_ref(node->batch);
         }
 
+        assert(snapshot->n_batches <= n_batches);
```
dvdhrm: Should be equal, shouldn't it?

teg: No, n_batches is based on n_gids, but there may not be policies for all gids.

src/launch/policy.c
teg: Fixed and pushed.

src/bus/policy.c
teg: Fixed.

meson.build
teg: Did not know that was possible. Thanks. Fixed.

src/launch/policy.c
teg: Ah, indeed. Good catch. Fixed.

src/launch/policy.c
teg: Yup. Done.

dbuch:
Isn't this code `https://github.com/bus1/dbus-broker/blob/master/src/launch/policy.c#L600` supposed to be removed as part of this pull request?

PR #72, opened 2017-10-19
Add SetPolicy() call to the controller

teg:
The broker allows names to be added/removed and the policy to be set for an existing listener. Together, this allows replacing all the configuration at run-time.

The launcher reloads the configuration, then removes all names that are no longer present, sets the new policy on the listener (which in turn updates the policy of each of the peers spawned from the listener). This can either be triggered by raising SIGHUP in the launcher, or by calling the internal ReloadConfig() call from the broker.

The broker now exposes the standard ReloadConfig() call on the org.freedesktop.DBus API which is forwarded to the launcher.

### DIFF
--- a/src/broker/controller-dbus.c
+++ b/src/broker/controller-dbus.c
@@ -302,6 +302,10 @@ static int controller_method_name_release(Controller *controller, const char *pa
         if (!name)
                 return CONTROLLER_E_NAME_NOT_FOUND;
 
+        r = controller_name_reset(name);
+        if (r)
+                return error_trace(r);
+
         controller_name_free(name);
 
         c_dvar_write(out_v, "()");
--- a/src/broker/controller-dbus.c
+++ b/src/broker/controller-dbus.c
@@ -507,12 +507,29 @@ static int controller_dispatch_object(Controller *controller, uint32_t serial, c
         return CONTROLLER_E_UNEXPECTED_PATH;
 }
 
+static int controller_dispatch_reply(Controller *controller, uint32_t serial, const char *signature, Message *message) {
+        ControllerReload *reload;
+        int r;
+
+        reload = controller_find_reload(controller, serial);
+        if (!reload)
+                return CONTROLLER_E_UNEXPECTED_REPLY;
+
+        if (strcmp(signature, ""))
+                return CONTROLLER_E_UNEXPECTED_SIGNATURE;
+
+        r = controller_reload_completed(reload);
+        if (r)
+                return error_trace(r);
+
+        controller_reload_free(reload);
+
+        return 0;
+}
+
 int controller_dbus_dispatch(Controller *controller, Message *message) {
         Connection *connection = &controller->connection;
         int r;
-
-        if (message->header->type != DBUS_MESSAGE_TYPE_METHOD_CALL)
-                return CONTROLLER_E_PROTOCOL_VIOLATION;
 
         r = message_parse_metadata(message);
         if (r > 0)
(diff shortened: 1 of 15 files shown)


src/broker/controller-dbus.c
```diff
@@ -505,28 +505,59 @@ static int controller_dispatch_object(Controller *controller, uint32_t serial, c
         return CONTROLLER_E_UNEXPECTED_PATH;
 }
 
+static int controller_dispatch_reply(Controller *controller, uint32_t serial, const char *signature, Message *message) {
+        ControllerReload *reload;
+        int r;
+
```
dvdhrm: Shouldn't we handle errors as well? If the launcher fails, it should be able to reply with an error? This call doesn't require it, but others might in the future. So once we generalize this, we should handle METHOD_RETURN and ERROR in some more generic sense. Fine for now, I think.

src/broker/controller-dbus.c
```diff
@@ -662,3 +693,42 @@ int controller_dbus_send_environment(Controller *controller, const char * const
...
+                     DBUS_MESSAGE_FIELD_INTERFACE, c_dvar_type_s, "org.bus1.DBus.Launcher",
+                     DBUS_MESSAGE_FIELD_MEMBER, c_dvar_type_s, "ReloadConfig");
+
+        r = c_dvar_end_write(&var, &data, &n_data);
+        if (r)
+                return error_origin(r);
+
+        r = message_new_outgoing(&message, data, n_data);
+        if (r)
+                return error_fold(r);
```
dvdhrm: This leaks `data`, doesn't it? Shouldn't it be marked as `_c_cleanup_(c_freep)` and NULL'ified after this call?

src/broker/controller.c
```diff
@@ -83,6 +83,56 @@ int controller_name_activate(ControllerName *name) {
...
+
+        slot = c_rbtree_find_slot(&controller->reload_tree, controller_reload_compare, &serial, &parent);
+        assert(slot);
+
+        reload = calloc(1, sizeof(*reload));
+        if (!reload)
+                return error_origin(-ENOMEM);
+
+        reload->controller = controller;
+        reload->controller_node = (CRBNode)C_RBNODE_INIT(reload->controller_node);
```
dvdhrm: Why no CONTROLLER_RELOAD_NULL?

src/broker/controller.c
```diff
@@ -83,6 +83,56 @@ int controller_name_activate(ControllerName *name) {
...
+        return NULL;
+}
+
+static int controller_reload_new(ControllerReload **reloadp, Controller *controller) {
+        CRBNode **slot, *parent;
+        ControllerReload *reload;
+        uint32_t serial = ++controller->serial;
+
+        slot = c_rbtree_find_slot(&controller->reload_tree, controller_reload_compare, &serial, &parent);
+        assert(slot);
```
dvdhrm: Ugh. I hate doing this on 32bit variables. Can't we make this a loop? Increment serial, look it up, continue if taken? Yeah yeah, trusted peer and so on... but still? Dunno..

src/bus/driver.c
```diff
@@ -1349,6 +1349,47 @@ static int driver_method_remove_match(Peer *peer, CDVar *in_v, uint32_t serial,
...
+
+                c_dvar_begin_write(&var, driver_type_out_unit, 1);
+                c_dvar_write(&var, "(");
+                driver_write_reply_header(&var, sender, reply_serial, driver_type_out_unit);
+                c_dvar_write(&var, "()");
+
+                r = driver_send_reply(sender, &var, reply_serial);
+                if (r)
+                        return error_trace(r);
+        }
```
dvdhrm: So you went with the single-object approach, meaning we never cleanup the pending reload-requests on disconnect, right?

Just for clarity: They're accounted on the uid, and pending on our trusted launcher. Hence, there is no exploit possible, IIUC.

src/bus/driver.c
```diff
@@ -1349,6 +1349,47 @@ static int driver_method_remove_match(Peer *peer, CDVar *in_v, uint32_t serial,
...
+
+        return 0;
+}
+
+static int driver_method_reload_config(Peer *peer, CDVar *in_v, uint32_t serial, CDVar *out_v) {
+        int r;
+
+        if (!peer_is_privileged(peer))
+                return DRIVER_E_PEER_NOT_PRIVILEGED;
```
dvdhrm: Did you check this with dbus-daemon?

test/dbus/util-broker.c
```diff
@@ -119,9 +119,39 @@ static int util_append_policy(sd_bus_message *m) {
         return 0;
 }
 
+static int util_method_reload_config(sd_bus_message *message, void *userdata, sd_bus_error *error) {
+        sd_bus *bus = sd_bus_message_get_bus(message);
```
dvdhrm: Urgs, we usually don't call non-inlined messages as part of initializations. In this case, it doesn't matter, I guess. But often it gets confusing if combined with cleanups.

src/broker/controller-dbus.c
teg: Correct. Kept it specific for now as the reply objects are anyway not generic.

src/broker/controller.c
teg: I agree, let's not rely on this being privileged.

src/broker/controller-dbus.c
teg: ugh... yeah, and this is wrong everywhere. Fixed it.

src/broker/controller.c
teg: Added.

src/bus/driver.c
teg: Correct. Though for the record, I don't see how cleaning up on disconnect would affect this regardless, as accounting is done per-uid and not per-connection.

test/dbus/util-broker.c
teg: Fair enough, fixed.

src/bus/driver.c
teg: Hm, I got it wrong. They allow anyone to reload config. Sounds ill-advised to me. For now, let's follow what they do, but something to discuss.

PR #78, opened 2017-11-16
Notify

teg:
Make dbus-broker Type=notify. See the first commit for justification.

### DIFF
--- a/src/launch/main.c
+++ b/src/launch/main.c
@@ -1134,6 +1134,8 @@ static int manager_reload_config(Manager *manager) {
         Service *service;
         int r;
 
+        (void) sd_notify(0, "RELOADING=1");
+
         c_rbtree_for_each_entry(service, &manager->services, rb)
                 service->state = SERVICE_STATE_DEFUNCT;
 
@@ -1156,6 +1158,8 @@ static int manager_reload_config(Manager *manager) {
         r = manager_add_services(manager);
         if (r)
                 return error_trace(r);
+
+        (void) sd_notify(0, "READY=1");
 
         return 0;
 }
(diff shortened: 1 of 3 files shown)


teg:
Just followed the recommendation in it man page. I agree, change it.

dvdhrm:
I pushed a fixed version of it. However, I did not copy your second patch. It was incomplete. It returned without sd_notify() in error_trace() calls. Those returns might return with positive error-codes, so they are not necessarily fatal. Hence, we cannot leave RELOADING=1 in those cases, but better do a READY=1.

Anyway, I also don't see the reason for STOPPING and RELOADING. Is there a particular purpose? Anyway. I'll close this now. Lets re-open a new PR if we want the second patch as well.

teg:
During reload we make blocking external calls, and are not fully
functional. If that somehow hangs it would be very useful to know (perhaps
PID1 should even consider this fatal). There is not really the potential
for hanging during stopping, but the principle remains: we are not fully
functional so might be useful for the parent to know.

On Nov 20, 2017 1:55 PM, "David Herrmann" <notifications@github.com> wrote:

I pushed a fixed version of it. However, I did not copy your second patch.
It was incomplete. It returned without sd_notify() in error_trace() calls.
Those returns might return with positive error-codes, so they are not
necessarily fatal. Hence, we cannot leave RELOADING=1 in those cases, but
better do a READY=1.

Anyway, I also don't see the reason for STOPPING and RELOADING. Is there a
particular purpose? Anyway. I'll close this now. Lets re-open a new PR if
we want the second patch as well.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/bus1/dbus-broker/pull/78#issuecomment-345687593>, or mute
the thread

PR #81, opened 2017-11-28
launcher: rework nss handling

teg:
The main feature of this PR is to use the correct UID when spawning transient units via systemd.

A few more nss fixes: we now cache the lookups, so we only resolve a user/group once during config/policy parsing. We immediately drop the cache though, so everything is re-resolved on reload.

We also now support user/group being the UID/GID directly, like the reference implementation always did. Lastly, we follow systemd and hard-code the "root" user/group to resolve to UID/GID 0 without going through NSS.

### DIFF
--- a/src/launch/main.c
+++ b/src/launch/main.c
@@ -633,6 +633,18 @@ static int manager_start_transient_unit(Manager *manager, Service *service) {
                         return error_origin(r);
 
                 if (service->user) {
+                        /*
+                         * Ideally we would unconditionally pass the UID
+                         * we are accounting on to systemd to run the service
+                         * under. However, in the case of the user instance,
+                         * systemd fails to start a transient unit if a user
+                         * is provided due to lack of permission. In practice
+                         * this works out ok, as in this case we would have
+                         * provided our own UID, which is systemd's UID, so
+                         * it would ammount to a no-op. It would have been
+                         * better if systemd could detect this case and not
+                         * fail, but in practice this is perfectly fine.
+                         */
                         r = sd_bus_message_open_container(method_call, 'r', "sv");
                         if (r < 0)
                                 return error_origin(r);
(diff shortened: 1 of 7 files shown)


src/launch/main.c
```diff
@@ -838,16 +843,23 @@ static int manager_load_service_file(Manager *manager, const char *path) {
...
+        if (user) {
+                r = nss_cache_get_uid(nss_cache, &uid, user);
+                if (r) {
+                        if (r == NSS_CACHE_E_INVALID_NAME) {
+                                fprintf(stderr, "Invalid user name '%s' in service file '%s'\n", user, path);
+                                r = 0;
+                                goto exit;
+                        }
+                }
+
```
dvdhrm: Weird empty line.

src/launch/main.c
```diff
@@ -838,16 +843,23 @@ static int manager_load_service_file(Manager *manager, const char *path) {
...
+        } else {
+                uid = getuid();
```
dvdhrm: Maybe a short comment here?

src/launch/nss-cache.c
```diff
@@ -0,0 +1,116 @@
...
+        node->rb = (CRBNode)C_RBNODE_INIT(node->rb);
+        memcpy(node->name, name, n_name + 1);
+
+        *nodep = node;
+        return 0;
+}
+
+static NSSCacheNode *nss_cache_node_free(NSSCacheNode *node) {
+        assert(!c_rbnode_is_linked(&node->rb));
+
```
dvdhrm: `if (!node) return;`?

src/launch/nss-cache.h
```diff
@@ -0,0 +1,36 @@
...
+#include <c-rbtree.h>
+#include <stdlib.h>
+#include <sys/types.h>
+
+typedef struct NSSCache NSSCache;
+
+enum {
+        _NSS_CACHE_E_SUCCESS,
+
+        NSS_CACHE_E_INVALID_NAME,
```
dvdhrm: "INVALID_NAME" sounds like the name was malformed, rather than just an unknown user.

PR #84, opened 2018-01-12
Tests

teg:
Some fixes/additions to the test suite and a fix for a bug it exposed.

### DIFF
--- a/test/dbus/test-driver.c
+++ b/test/dbus/test-driver.c
@@ -4,7 +4,9 @@
 
 #include <c-macro.h>
 #include <stdlib.h>
-#include "../../src/dbus/protocol.h"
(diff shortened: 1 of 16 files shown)


test/dbus/test-driver.c
```diff
@@ -568,7 +570,8 @@ static void test_name_has_owner(void) {
         {
                 _c_cleanup_(sd_bus_flush_close_unrefp) sd_bus *bus = NULL;
                 _c_cleanup_(sd_bus_message_unrefp) sd_bus_message *reply = NULL;
-                const char *unique_name, *owned;
+                const char *unique_name;
+                bool owned;
```
dvdhrm: This must be `int`.

PR #88, opened 2018-01-19
NameOwnerChanged subscriptions

teg:
Make NameOwnerChanged matches work like subscriptions on  the name in question, rather than driver-wide matches. This should  greatly improve scalability of connect/disconnect as well as regular name owner change.

### DIFF
--- a/src/bus/peer.c
+++ b/src/bus/peer.c
@@ -197,7 +197,7 @@ int peer_new_with_fd(Peer **peerp,
         peer->charges[1] = (UserCharge)USER_CHARGE_INIT;
         peer->charges[2] = (UserCharge)USER_CHARGE_INIT;
         peer->owned_names = (NameOwner)NAME_OWNER_INIT;
-        peer->matches = (MatchRegistry)MATCH_REGISTRY_INIT(peer->matches);
+        peer->sender_matches = (MatchRegistry)MATCH_REGISTRY_INIT(peer->sender_matches);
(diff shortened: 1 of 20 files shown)


src/bus/peer.c
```diff
@@ -449,6 +502,26 @@ int peer_add_match(Peer *peer, const char *rule_string) {
...
+         *  - if it is a sender match on the name.
+         */
+        if (strcmp(rule->keys.sender, "org.freedesktop.DBus") == 0) {
+                if (rule->keys.filter.member && strcmp(rule->keys.filter.member, "NameOwnerChanged") == 0 &&
+                    rule->keys.filter.args[0] && strcmp(rule->keys.filter.args[0], "org.freedesktop.DBus") != 0 &&
+                    rule->keys.filter.args[0][0] != ':')
+                        return c_container_of(rule->registry, Name, name_owner_changed_matches);
+        } else if (rule->keys.sender[0] != ':') {
+                return c_container_of(rule->registry, Name, sender_matches);
+        }
```
dvdhrm: Yeah... This needs a big fat `XXX: WTF?`

I am fine with it for now, since we had the same conceptual issue before. But this is not cool. Maybe add a field to `rule` that pins the name. Or a boolean at least.

dvdhrm:
Very nice! I love it!

The rule-to-name issue is still ugly, but an unrelated issue. So lets fix it later.

All merged!

PR #90, opened 2018-01-23
test/util-broker/spawn: add a barrier before returning

teg:
Make sure the broker/daemon is fully operational by the time spawn() returns. Without
this the setup of the broker could still be happening when we start the performance
measurements.


test/dbus/util-broker.c
```diff
@@ -174,7 +174,6 @@ void util_fork_broker(sd_bus **busp, sd_event *event, int listener_fd, pid_t *pi
 
                 r = execl("./src/dbus-broker",
                           "./src/dbus-broker",
-                          "--verbose",
```
dvdhrm: Is that in any way related?

teg: No. Split out now.

PR #94, opened 2018-02-01
launch/policy: improve warnings

teg:
Improve warnings emitted by the launcher on deprecated policy. Make sure we always log when ignoring policy that is not a noop, but do not log about ignoring policy that would not have any effect.

This also fixes an (obscure) bug related to eavesdrop policies.


dvdhrm:
For completeness:

PR #95, opened 2018-02-05
bus/log: include more meta information when policy denies transmission

teg:


src/bus/bus.c
```diff
@@ -107,17 +107,55 @@ static int bus_log_commit_policy(Bus *bus, const char *action, uint64_t sender_i
...
+                        c_rbtree_for_each_entry(ownership,
+                                                &receiver_names->owner->ownership_tree,
+                                                owner_node)
+                                log_appendf(log,
+                                            "DBUS_BROKER_RECEIVER_WELL_KNOWN_NAME_%zu=%s\n",
+                                            i++, ownership->name->name);
+                } else if (receiver_names->type == NAME_SET_TYPE_SNAPSHOT) {
+                        for (size_t i = 0; i < receiver_names->snapshot->n_names; ++i)
+                                log_appendf(log,
+                                            "DBUS_BROKER_SENDER_WELL_KNOWN_NAME_%zu=%s\n",
```
dvdhrm: This must be `DBUS_BROKER_RECEIVER_WELL_KNOWN_NAME_`

src/bus/bus.c
```diff
@@ -81,14 +81,25 @@ Peer *bus_find_peer_by_name(Bus *bus, Name **namep, const char *name_str) {
...
+        if (policy_type)
+                log_appendf(log, "DBUS_BROKER_POLICY_TYPE=%s\n", policy_type);
+
+        if (sender_label)
+                log_appendf(log, "DBUS_BROKER_SENDER_SECURITY_LABEL=%s\n",
+                            sender_label);
+
+        if (receiver_label)
+                log_appendf(log, "DBUS_BROKER_RECEIVER_SECURITY_LABEL=%s\n",
+                            receiver_label);
```
dvdhrm: So we finally treat security labels as strings? I am not sure what our last agreement there was. We always tried avoiding this, but it might just not be worth it. Anyway, just to be clear: From now on we follow what selinux does and treat the security-label-output as a zero-terminated string (without any internal zeros)?

src/bus/bus.c
```diff
@@ -153,9 +164,13 @@ static int bus_log_commit_policy(Bus *bus, const char *action, uint64_t sender_i
 }
 
 int bus_log_commit_policy_send(Bus *bus, uint64_t sender_id, uint64_t receiver_id, NameSet *sender_names, NameSet *receiver_names, Message *message) {
-        return bus_log_commit_policy(bus, "send", sender_id, receiver_id, sender_names, receiver_names, message);
+        return bus_log_commit_policy(bus, "send", "internal", NULL, NULL, sender_id, receiver_id, sender_names, receiver_names, message);
 }
 
 int bus_log_commit_policy_receive(Bus *bus, uint64_t receiver_id, uint64_t sender_id, NameSet *sender_names, NameSet *receiver_names, Message *message) {
-        return bus_log_commit_policy(bus, "receive", sender_id, receiver_id, sender_names, receiver_names, message);
+        return bus_log_commit_policy(bus, "receive", "internal", NULL, NULL, sender_id, receiver_id, sender_names, receiver_names, message);
```
dvdhrm: TODO item: We could also distinguish between implicit builtin rules and xml rules.

src/bus/peer.c
```diff
@@ -651,9 +651,12 @@ int peer_queue_call(PolicySnapshot *sender_policy, NameSet *sender_names, MatchR
...
                                        message->header->type);
         if (r) {
-                if (r == POLICY_E_ACCESS_DENIED) {
+                if (r == POLICY_E_ACCESS_DENIED || r == POLICY_E_SELINUX_ACCESS_DENIED) {
                         log_append_here(receiver->bus->log, LOG_WARNING, 0);
-                        r = bus_log_commit_policy_send(receiver->bus, sender_id, receiver->id, sender_names, &receiver_names, message);
+                        if (r == POLICY_E_ACCESS_DENIED)
+                                r = bus_log_commit_policy_send(receiver->bus, sender_id, receiver->id, sender_names, &receiver_names, message);
+                        else
+                                r = bus_log_commit_policy_selinux_send(receiver->bus, sender_policy->seclabel, receiver->policy->seclabel, sender_id, receiver->id, sender_names, &receiver_names, message);
```
dvdhrm: Thinking about it more: Can't we make the "policy-type" argument an enum? Rather htan making it separate function names... We can still store strings in the log to avoid log-ABI. But the internal API now looks weird.

src/bus/bus.c
teg: As far as I'm aware we are guaranteed no internal zeros. Otherwise it could be garbage, but just for logging I think that's something we can deal with if it happens

src/bus/bus.c
teg: Agreed

src/bus/bus.c
teg: Will fix. Thanks!

src/bus/peer.c
teg: Done.

PR #97, opened 2018-02-15
launch: support supplementing the at_console users at compile time

teg:
This is WIP, and so far only the changes to the controller API has been implemented.

### DIFF
--- a/src/launch/policy.c
+++ b/src/launch/policy.c
@@ -117,6 +117,13 @@ static void policy_entries_deinit(PolicyEntries *entries) {
                 policy_record_free(record);
 }
 
+static bool policy_entries_is_empty(PolicyEntries *entries) {
+        return (c_list_is_empty(&entries->connect_list) &&
+                c_list_is_empty(&entries->own_list) &&
+                c_list_is_empty(&entries->send_list) &&
+                c_list_is_empty(&entries->recv_list));
+}
+
 static int policy_node_compare(CRBTree *t, void *k, CRBNode *n) {
         PolicyNode *node = c_container_of(n, PolicyNode, policy_node);
         uint32_t uidgid = *(uint32_t*)k;
@@ -734,17 +741,11 @@ static void policy_optimize_trim(Policy *policy) {
          */
 
         c_rbtree_for_each_entry_safe(node, t_node, &policy->uid_tree, policy_node)
-                if (c_list_is_empty(&node->entries.connect_list) &&
-                    c_list_is_empty(&node->entries.own_list) &&
-                    c_list_is_empty(&node->entries.send_list) &&
-                    c_list_is_empty(&node->entries.recv_list))
+                if (policy_entries_is_empty(&node->entries))
                         policy_node_free(node);
 
         c_rbtree_for_each_entry_safe(node, t_node, &policy->gid_tree, policy_node)
-                if (c_list_is_empty(&node->entries.connect_list) &&
-                    c_list_is_empty(&node->entries.own_list) &&
-                    c_list_is_empty(&node->entries.send_list) &&
-                    c_list_is_empty(&node->entries.recv_list))
+                if (policy_entries_is_empty(&node->entries))
                         policy_node_free(node);
 }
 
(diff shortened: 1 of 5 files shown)


dvdhrm:
I like it! Haven't done a full review yet, but I like the overall structure!

src/launch/policy.c
```diff
@@ -995,6 +962,60 @@ int policy_export(Policy *policy, sd_bus_message *m) {
                         return error_origin(r);
         }
 
+        r = sd_bus_message_open_container(m, 'r', "uu(" POLICY_T_BATCH ")");
+        if (r < 0)
+                return error_origin(r);
+
+        r = sd_bus_message_append(m, "uu", 0, SYSTEMUIDMAX);
```
dvdhrm: `r = sd_bus_message_append(m, "uu", UINT32_C(0), UINT32_C(SYSTEMUIDMAX));`

src/launch/policy.c
```diff
@@ -995,6 +962,60 @@ int policy_export(Policy *policy, sd_bus_message *m) {
...
+
+        r = sd_bus_message_close_container(m);
+        if (r < 0)
+                return error_origin(r);
+
+        r = sd_bus_message_open_container(m, 'r', "uu(" POLICY_T_BATCH ")");
+        if (r < 0)
+                return error_origin(r);
+
+        r = sd_bus_message_append(m, "uu", SYSTEMUIDMAX + 1, -1);
```
dvdhrm: Wrong types. Same problem as above.

src/launch/policy.c
```diff
@@ -886,10 +886,57 @@ static int policy_export_xmit(Policy *policy, CList *list1, CList *list2, sd_bus
                 "a(buu(" POLICY_T_BATCH "))"                                    \
                 "a(ss)"
 
+static int policy_export_console(Policy *policy, sd_bus_message *m, PolicyEntries *entries, uid_t uid_start, uid_t uid_end) {
+        int r;
+
+        if (uid_start >= uid_end)
+                return 0;
```
dvdhrm: It is really not obvious that uid_end is exclusive. Can we label it as such? uid_end_exclusive? Or better: uid_start and n_uids?

src/launch/policy.c
```diff
@@ -886,10 +886,57 @@ static int policy_export_xmit(Policy *policy, CList *list1, CList *list2, sd_bus
                 "a(buu(" POLICY_T_BATCH "))"                                    \
                 "a(ss)"
 
+static int policy_export_console(Policy *policy, sd_bus_message *m, PolicyEntries *entries, uid_t uid_start, uid_t uid_end) {
```
dvdhrm: Why uid_t everywhere in this file? I'd prefer if the caller does the conversion. We deal with uint32_t everywhere else here.

src/launch/policy.c
```diff
@@ -999,64 +1046,42 @@ int policy_export(Policy *policy, sd_bus_message *m) {
...
-        r = sd_bus_message_open_container(m, 'r', "buu(" POLICY_T_BATCH ")");
-        if (r < 0)
-                return error_origin(r);
+        {
+                uid_t prev_uid = 0;
 
-        r = sd_bus_message_append(m, "buu", false, 0, SYSTEMUIDMAX);
-        if (r < 0)
-                return error_origin(r);
+                if (n_at_console_uids > 0) {
```
dvdhrm: Stupid qsort() does not work with NULL+0, right? Meh...

src/launch/policy.c
```diff
@@ -999,64 +1046,42 @@ int policy_export(Policy *policy, sd_bus_message *m) {
...
-        if (r < 0)
-                return error_origin(r);
+                        r = policy_export_console(policy, m, &policy->at_console_entries, at_console_uids[i], at_console_uids[i] + 1);
+                        if (r < 0)
+                                return error_trace(r);
 
-        r = sd_bus_message_open_container(m, 'r', "buu(" POLICY_T_BATCH ")");
-        if (r < 0)
-                return error_origin(r);
+                        prev_uid = at_console_uids[i];
```
dvdhrm: We could coalesce consequetive entries here. But... eh... not worth it?

src/launch/policy.c
```diff
@@ -999,64 +1046,42 @@ int policy_export(Policy *policy, sd_bus_message *m) {
...
+                        qsort(at_console_uids, n_at_console_uids, sizeof(*at_console_uids), policy_uid_compare);
 
-        r = policy_export_connect(policy, NULL, &policy->no_console_entries.connect_list, m);
-        r = r ?: policy_export_own(policy, NULL, &policy->no_console_entries.own_list, m);
-        r = r ?: policy_export_xmit(policy, NULL, &policy->no_console_entries.send_list, m);
-        r = r ?: policy_export_xmit(policy, NULL, &policy->no_console_entries.recv_list, m);
-        if (r)
-                return error_trace(r);
+                        while (at_console_uids[n_at_console_uids - 1] > SYSTEMUIDMAX)
+                                --n_at_console_uids;
```
dvdhrm: This fails if all uids are > SYSTEMUIDMAX. That is, this loop will happily read beyond borders after clearing the last uid. You need something like:

```
while (n_at_console_uids && at_console_uids[n_at_console_uids - 1] > SYSTEMUIDMAX)
```

src/launch/policy.c
teg: Fixed.

src/launch/policy.c
teg: Fixed.

src/launch/policy.c
teg: Fixed with `n_uids`.

src/launch/policy.c
teg: Done.

src/launch/policy.c
teg: Fixed.

src/launch/policy.c
teg: My thoughts exactly.

src/launch/policy.c
teg: Agreed, fixed.

src/launch/policy.c
teg: As a wise man (you) pointed out in private chat, NULL+0 is actually fine: https://sourceware.org/git/?p=glibc.git;a=blob;f=stdlib/qsort.c;h=264a06b8a924a1627b3c0fd507a3e2ca38dbc8a0;hb=HEAD#l98 I'll simplify.

dvdhrm:
The uid-array we pass does not prevent duplicates. For instance, there might be two sets with `uid==-1`. I just checked, and what we end up doing is merging them. I think this is the right thing to do. I just felt like mentioning it. Not sure whether you thought of that.

teg:
Yes, I did, and yes, I agree it is right.

PR #98, opened 2018-02-19
launch/config: Make correct path for included file

ylse:
When config file has a include node with relative path(e.g. <include>
some.conf</include>), it makes a wrong path. The included file path
should be in the same directory as config file(denoted as prefix).

Therefore add codes to judge whether prefix is a file or directory here.
If prefix is a directory, included file's name is added after prefix.
If prefix is a file, config file's name is replaced by included file's
name. So that it can create correct path for included file.

### DIFF
--- a/src/launch/config.c
+++ b/src/launch/config.c
@@ -6,6 +6,7 @@
 #include <c-macro.h>
 #include <expat.h>
 #include <stdlib.h>
+#include <sys/stat.h>
 #include "dbus/protocol.h"
 #include "launch/config.h"
 #include "launch/nss-cache.h"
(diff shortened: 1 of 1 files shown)


dvdhrm:
Thanks for the patch! I think it could be reduced to this, though:

```
diff --git a/src/launch/config.c b/src/launch/config.c
index 6a68699..579471b 100644
--- a/src/launch/config.c
+++ b/src/launch/config.c
@@ -1022,7 +1022,7 @@ static void config_parser_end_fn(void *userdata, const XML_Char *name) {
                                     state->file,
                                     state->current->include.selinux_root_relative ?
                                         bus_selinux_policy_root() :
-                                        state->file->path,
+                                        NULL,
                                     state->current->cdata);
                 if (r) {
                         state->error = error_trace(r);
```

This bug was caused by our selinux conversion. See dd2923eeb9fac3acce3fbf856ea6745c1624861e.

dvdhrm:
I will close this in favor of #101. Lets continue the discussion there.

Thanks for the report!

PR #99, opened 2018-02-19
launch/config: fix path computation for <include>

### LINKED ISSUE #96: Configuration parse logic is including the top level conf file
Configuration of "`/usr/share/dbus-1/system.conf`" is looking like this for us

```
...
...
<include>addon.conf</include>
...
...
```

dbus-broker is complaining about an unknown file @ "`/usr/share/dbus-1/system.conf/addon.conf`".

Bug found by Yang.

dvdhrm:
This is a shorter fix for #98, supposed to fix #96. I tested it locally, and it works as expected.

@ylse: If you provide a `Reviewed-by: Name [<email>]` line, I can append it to the patch.

### DIFF
--- a/src/launch/config.c
+++ b/src/launch/config.c
@@ -1022,7 +1022,7 @@ static void config_parser_end_fn(void *userdata, const XML_Char *name) {
                                     state->file,
                                     state->current->include.selinux_root_relative ?
                                         bus_selinux_policy_root() :
-                                        state->file->path,
+                                        NULL,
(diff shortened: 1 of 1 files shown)


ylse:
I have verified the change and it works well. Thanks a lot.
Reviewed-by: Yang [lylyly86006869@126.com]

dvdhrm:
Thanks a lot! Amended and pushed.

PR #100, opened 2018-02-19
launch: scope rework

teg:
Rework the scope logic to make it more fine-grained and make it possible to use the launcher in other settings than as the user/system bus.

In particular, the --scope variable now only indicates which bus the corresponding systemd instance can be found on, as well as some defaults for other configuration options.

### DIFF
--- a/src/launch/config.c
+++ b/src/launch/config.c
@@ -142,6 +142,9 @@ ConfigNode *config_node_free(ConfigNode *node) {
         case CONFIG_NODE_INCLUDE:
                 config_path_unref(node->include.file);
                 break;
+        case CONFIG_NODE_SERVICEDIR:
+                free(node->servicedir.path);
+                break;
         case CONFIG_NODE_LIMIT:
                 free(node->limit.name);
                 break;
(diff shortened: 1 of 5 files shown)


src/launch/main.c
```diff
@@ -1532,7 +1540,7 @@ static int run(void) {
                 return MAIN_FAILED;
         }
 
-        r = manager_run(manager, !strcmp(main_arg_scope, "system"));
+        r = manager_run(manager, main_arg_audit);
```
dvdhrm: No reason to propagate that value, anymore, right? `manager_run_child()` should use it directly, imho. We always make sure our commandline arguments are constant globals, so no harm, I think.

src/launch/config.c
```diff
@@ -781,6 +790,8 @@ static void config_parser_begin_fn(void *userdata, const XML_Char *name, const X
                 if (r)
                         goto failed;
 
+                c_list_link_tail(&state->root->servicedir_list, &node->servicedir_link);
+
```
dvdhrm: I don't think this works. We don't parse configurations in linear order. The only guarantee you get is that the node-list *afterwards* is in linear order.

For example, we always parse a single file *fully* before parsing any included files. But we make sure when parsing the included files, we insert them at the correct position in the node-tree/list. But if you now keep a separate order in `servicedir_list`, based on the time we parse nodes, you will get out-of-sync.

I think the only valid solution is to do a post-processing step. Once all the nodes have been parsed, go over them *once more* linearly, to get further links (like the servicedir list).

src/launch/main.c
```diff
@@ -69,8 +69,8 @@ static bool             main_arg_audit = false;
 static const char *     main_arg_broker = BINDIR "/dbus-broker";
 static bool             main_arg_force = false;
 static const char *     main_arg_listen = NULL;
-static const char *     main_arg_scope = "system";
 static const char *     main_arg_policypath = NULL;
+static bool             main_arg_user_scope = false;
```
dvdhrm: Mh. I understand having a string is cumbersome, but a boolean reads weird as well. Mhh. Technically this should be an enum. Stupid C.

dvdhrm: Wow, how did this comment end up in this commit? *sigh*

dvdhrm:
Merged!

PR #107, opened 2018-02-22
Solicited directed signals

teg:
Treat the inability to receive solicited directed signals as fatal, and disconnect the receiving peer.

### DIFF
--- a/src/bus/peer.c
+++ b/src/bus/peer.c
@@ -595,106 +595,6 @@ void peer_flush_matches(Peer *peer) {
         }
 }
 
-int peer_queue_call(PolicySnapshot *sender_policy, NameSet *sender_names, ReplyOwner *sender_replies, User *sender_user, uint64_t sender_id, Peer *receiver, Message *message) {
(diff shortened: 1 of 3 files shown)


src/bus/peer.c
```diff
@@ -794,10 +808,16 @@ int peer_queue_call(PolicySnapshot *sender_policy, NameSet *sender_names, ReplyO
 
         r = connection_queue(&receiver->connection, sender_user, message);
         if (r) {
-                if (CONNECTION_E_QUOTA)
-                        return PEER_E_QUOTA;
-                else
+                if (CONNECTION_E_QUOTA) {
+                        if (message->header->type == DBUS_MESSAGE_TYPE_SIGNAL &&
+                            peer_signal_is_solicited(receiver, sender_id, message)) {
+                                connection_shutdown(&receiver->connection);
```
dvdhrm: Maybe rename the function to `peer_message_is_solicited()` and check for signal-type there? Looks weird to have this type-check at a random place.

PR #112, opened 2018-03-01
launch/transient: namespace transient units by launcher instance

teg:
Now that we explicitly support several launcher instances per scope
(i.e., systemd instance), we need to further namespace the service
names we generate ourselves.

Each launcher instance has a unique name on the bus corresponding
to the scope, and systemd binds the transient unit to this name.
The new naming scheme becomes:

  `dbus-<launcher unique name>-<service name>@<instance number>.service`

### DIFF
--- a/src/launch/main.c
+++ b/src/launch/main.c
@@ -461,11 +461,16 @@ static int manager_start_unit(Manager *manager, Service *service) {
 static int manager_start_transient_unit(Manager *manager, Service *service) {
         _c_cleanup_(sd_bus_message_unrefp) sd_bus_message *method_call = NULL;
         _c_cleanup_(c_freep) char *unit = NULL;
+        const char *unique_name;
         int r;
 
         service->slot = sd_bus_slot_unref(service->slot);
 
-        r = asprintf(&unit, "dbus-%s@%"PRIu64".service", service->name, service->instance++);
+        r = sd_bus_get_unique_name(manager->bus_regular, &unique_name);
+        if (r < 0)
+                return error_origin(r);
+
+        r = asprintf(&unit, "dbus-%s-%s@%"PRIu64".service", unique_name, service->name, service->instance++);
         if (r < 0)
                 return error_origin(-errno);
 


src/launch/main.c
```diff
@@ -461,11 +461,16 @@ static int manager_start_unit(Manager *manager, Service *service) {
...
         _c_cleanup_(c_freep) char *unit = NULL;
+        const char *unique_name;
         int r;
 
         service->slot = sd_bus_slot_unref(service->slot);
 
-        r = asprintf(&unit, "dbus-%s@%"PRIu64".service", service->name, service->instance++);
+        r = sd_bus_get_unique_name(manager->bus_regular, &unique_name);
+        if (r < 0)
+                return error_origin(r);
```
dvdhrm: `sd-bus` caches the unique-name, but this call still blocks in case you connect asynchronously. I recommend calling this directly after sd_bus_open_{system,user}() and discard the value immediately. This way, we make the connect-attempt synchronous, and avoid the entire async-but-randomly-blocking-setup of sd-bus. This is especially important when we already sd_notify'ed pid1.

src/launch/main.c
```diff
@@ -461,11 +461,16 @@ static int manager_start_unit(Manager *manager, Service *service) {
...
+
+        r = asprintf(&unit, "dbus-%s-%s@%"PRIu64".service", unique_name, service->name, service->instance++);
```
dvdhrm: Apart from the length restriction, the namespacing here is fine. Just checked with unit-name.[ch] in systemd.

teg: Thanks!

src/launch/main.c
teg: I see two possible ideal scenarios:

 * Make connect synchronous, before sd_notify'ing pid1. This would be trivial to do, but would mean delaying boot across the board, probably unnecessarily.
 * Keep everything fully async, and don't change when we sd_notify. However, only start processing requests from the broker after the connection to the regular bus has completed successfully. This would mean that we always have the unique name available, so we never block after sd_notify'ing and that we never reorder the incoming requests.

The latter is for all intents and purposes functionally equivalent to the current patch.

PR #113, opened 2018-03-21
build: make documentation optional

teg:
Make building documentation opt-in.

See: <http://lists.openembedded.org/pipermail/openembedded-devel/2018-March/117419.html>.

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -1,5 +1,6 @@
 option('audit', type: 'boolean', value: false, description: 'Audit support')
 option('launcher', type: 'boolean', value: true, description: 'Build compatibility launcher')
+option('docs', type: 'boolean', value: false, description: 'Build documentation')
 option('reference-test', type: 'boolean', value: false, description: 'Run test suite against reference implementation')
 option('selinux', type: 'boolean', value: false, description: 'SELinux support')
 option('system-console-users', type: 'array', value: [], description: 'Additional set of names of system-users to be considered at-console')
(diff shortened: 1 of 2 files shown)


meson_options.txt
```diff
@@ -1,5 +1,6 @@
 option('audit', type: 'boolean', value: false, description: 'Audit support')
 option('launcher', type: 'boolean', value: true, description: 'Build compatibility launcher')
+option('docs', type: 'boolean', value: false, description: 'Build documentation')
```
dnicolodi: Would't it be better to have a `true` default value, at least for compatibility with the old build defaults?

dvdhrm: Cannot speak for Tom here, but our preference is generally to default to 'false'. I know we didn't follow this with the 'launcher' option, but maybe we should?

We also document those changes in 'NEWS', so package maintainers can adapt to it. I understand, though, that we should keep burden on them as small as possible.

Tom, any specific reason to default to false, other than that it feels right that way?

teg: Sorry for the late reply.

My reasoning was the same as David's, we should default to false for anything that isn't strictly speaking necessary on all systems.

As for the launcher, I'm fine with that being enabled by default, at least until we document the broker API, declare it stable and someone actually uses it directly (making it usable without the launcher).

dnicolodi: No strong preference. I was pointing this out in the case that there were no strong reasons behind the choice, but the justifications to default to `false` is sound.

PR #114, opened 2018-04-09
Fix typos

dnicolodi:
Just two typos spotted by `lintian` in the Debian package.


dvdhrm:
Both look good, thanks!

PR #120, opened 2018-05-02
match: index by path/interface/member in addition to destination

teg:
This should all but eliminate false-positives when iterating potential matches for a signal.

In a follow-up, we will ref-count identical matches, to further reduce the number of matches we have to linearly iterate through.

### DIFF
--- a/src/bus/match.c
+++ b/src/bus/match.c
@@ -437,7 +437,7 @@ static MatchRegistryByInterface *match_registry_by_interface_unref(MatchRegistry
         if (!registry || --registry->n_refs > 0)
                 return NULL;
 
-        assert(c_list_is_empty(&registry->rule_list));
+        assert(c_rbtree_is_empty(&registry->member_tree));
 
         c_rbnode_unlink(&registry->registry_node);
         match_registry_by_path_unref(registry->registry_by_path);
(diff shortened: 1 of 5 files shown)


src/bus/match.c
```diff
@@ -580,16 +693,23 @@ void match_registry_init(MatchRegistry *registry) {
...
  * mach_registry_flush() - XXX
  */
 void match_registry_flush(MatchRegistry *registry) {
-        MatchRule *rule, *rule_safe;
+        MatchRegistryByPath *registry_by_path, *registry_by_path_safe;
 
-        c_list_for_each_entry_safe(rule, rule_safe, &registry->subscription_list, registry_link)
-                match_rule_unlink(rule);
+        c_rbtree_for_each_entry_safe_postorder(registry_by_path, registry_by_path_safe, &registry->subscription_tree, registry_node)
+                match_registry_by_path_flush(registry_by_path);
```
dvdhrm: This has to be:

```
match_registry_by_path_ref(registry_by_path);
match_registry_by_path_flush(registry_by_path);
match_registry_by_path_unref(registry_by_path);
```

src/bus/match.c
```diff
@@ -339,6 +339,62 @@ static bool match_keys_match_filter(MatchKeys *keys, MatchFilter *filter) {
         return true;
 }
 
+static int match_registry_by_path_compare(CRBTree *tree, void *k, CRBNode *rb) {
+        MatchRegistryByPath *registry = c_container_of(rb, MatchRegistryByPath, registry_node);
+        const char *path1 = k ?: "", *path2 = registry->path;
```
dvdhrm: Mh. This is hidden deep down. Shouldn't the caller of the lookup pick the empty string rather than NULL?

src/bus/match.c
```diff
@@ -488,20 +574,47 @@ bool match_rule_match_filter(MatchRule *rule, MatchFilter *filter) {
...
                         return rule;
 
         return NULL;
 }
 
-MatchRule *match_rule_next_monitor_match(MatchRegistry *registry, MatchRule *rule, MatchFilter *filter) {
-        c_list_for_each_entry_continue(rule, &registry->monitor_list, registry_link)
-                if (match_rule_match_filter(rule, filter))
-                        return rule;
+MatchRule *match_rule_next_match(CRBTree *tree, MatchRule *rule, MatchFilter *filter) {
```
dvdhrm: static?

src/bus/match.c
teg: As discussed, leaving this as is for now.

src/bus/match.c
teg: Fixed.

src/bus/match.c
teg: Fixed.

src/bus/match.c
```diff
@@ -488,20 +574,47 @@ bool match_rule_match_filter(MatchRule *rule, MatchFilter *filter) {
         return match_keys_match_filter(&rule->keys, filter);
 }
 
-MatchRule *match_rule_next_subscription_match(MatchRegistry *registry, MatchRule *rule, MatchFilter *filter) {
-        c_list_for_each_entry_continue(rule, &registry->subscription_list, registry_link)
+MatchRule *match_rule_next_match_by_path(MatchRegistryByPath *registry, MatchRule *rule, MatchFilter *filter) {
```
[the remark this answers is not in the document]
teg: Fixed.

PR #122, opened 2018-05-09
driver: add Peer interface

teg:
This is required by the spec. Also, allow method calls with an empty destination.


teg:
I added test-cases to make sure that this behaves the same on dbus-daemon and dbus-broker.

PR #124, opened 2018-05-11
driver: implement the standard Properties interface

teg:
The driver should expose two properties, Features and Interfaces. Implement
Get and GetAll.


teg:
Rebased and added Set method (which always fails with PropertyReadOnly).

PR #125, opened 2018-05-14
Registration

teg:
Fix the error handling related to the Hello() call.


src/bus/driver.c
```diff
@@ -1728,47 +1734,70 @@ static int driver_handle_method(const DriverMethod *method, Peer *peer, const ch
...
+        { "GetId",                                      NULL,                           driver_method_get_id,                                           c_dvar_type_unit,       driver_type_out_s },
+        { "Introspect",                                 NULL,                           driver_method_introspect,                                       c_dvar_type_unit,       driver_type_out_s },
+        { "BecomeMonitor",                              "/org/freedesktop/DBus",        driver_method_become_monitor,                                   driver_type_in_asu,     driver_type_out_unit },
+        { "Ping",                                       NULL,                           driver_method_ping,                                             c_dvar_type_unit,       driver_type_out_unit },
+        { "GetMachineId",                               NULL,                           driver_method_get_machine_id,                                   c_dvar_type_unit,       driver_type_out_s },
+        { },
```
dvdhrm: Errr, doesn't this mean *all* these methods are now dispatched on "org.freedesktop.DBus"? Ping, GetMachineId, etc. should not be available on the generic interface, should they?

teg: You are right. That's an oversight. Will push fix.

teg:
Pushed rebase with requested fix.

dvdhrm:
Looks good!

PR #127, opened 2018-05-15
broadcasts: split collecting the notifications from sending the messages

teg:


dvdhrm:
Can you respin this on master?

PR #129, opened 2018-05-16
Extend NSS Cache

dvdhrm:
Hey

This is in preparation for the `<user>%</user>` attribute of the XML config. This does not fully implement it, but only contains the preparations in the NSS cache.

The only undiscussed change here is the re-introduction of the 'root' group as pre-populated entry. Given that we have to pre-propulate the full "struct passwd" now, this means that we must also know the "pw->pw_gid" entry. Lets just define this as 0 and reintroduce the 'root' group as group 0. This is universally true and should be fine.

Thanks
David


dvdhrm:
Pushed a fix for a `(const char *) -> (char *)` cast.

PR #136, opened 2018-05-17
drop privileges in launcher

teg:
We already spawn the broker with reduced privileges, drop the privileges in the launcher too.


dvdhrm:
Gah, it should be called 'drop_privileges'... I messed that up originally. 'drop_permissions' makes no sense, right?

PR #142, opened 2018-05-24
log: add log messages on quota failures and related disconnects

teg:
Make sure we log about all quota failures. Also, log about failing transactions due to quota failure (separately). Together this should make debugging considerably simpler.

There are still cases where clients are disconnected which are not logged, namely where the peer has exceeded its own quota or attempts a protocol violation. These should be addressed separately, as they are more complicated to get right, and less likely to crop up in real life.

Calling org.freedesktop.DBus.GetId on org.freedesktop.DBus in a loop without reading the replies now logs about the resulting disconnect as follows:

```
UID 1000 exceeded its 'bytes' quota on UID 1000.
Peer :1.107 is being disconnected as it does not have the resources to receive a reply or unicast signal it expects.
```

### DIFF
--- a/src/bus/driver.c
+++ b/src/bus/driver.c
@@ -1976,7 +1976,7 @@ static int driver_dispatch_interface(Peer *peer, uint32_t serial, const char *in
                         log_append_here(peer->bus->log, LOG_WARNING, 0);
                         bus_log_append_policy_send(peer->bus,
                                                    (r == POLICY_E_ACCESS_DENIED ? BUS_LOG_POLICY_TYPE_INTERNAL : BUS_LOG_POLICY_TYPE_SELINUX),
-                                                   peer->id, ADDRESS_ID_INVALID, &names, NULL, peer->policy->seclabel, NULL, message);
+                                                   peer->id, ADDRESS_ID_INVALID, &names, NULL, peer->policy->seclabel, peer->bus->seclabel, message);
                         r = log_commitf(peer->bus->log, "A security policy denied :1.%llu to send method call %s:%s.%s to org.freedesktop.DBus.",
                                         peer->id, path, interface, member);
                         if (r)
--- a/src/bus/driver.c
+++ b/src/bus/driver.c
@@ -348,10 +348,22 @@ static int driver_monitor(Bus *bus, Peer *sender, Message *message) {
 
                 r = connection_queue(&receiver->connection, NULL, message);
                 if (r) {
-                        if (r == CONNECTION_E_QUOTA)
+                        if (r == CONNECTION_E_QUOTA) {
+                                NameSet sender_names = NAME_SET_INIT_FROM_OWNER(sender ? &sender->owned_names : NULL);
+
                                 connection_shutdown(&receiver->connection);
-                        else
+
+                                log_append_here(bus->log, LOG_WARNING, 0);
+                                bus_log_append_transaction(bus, sender ? sender->id : ADDRESS_ID_INVALID, receiver->id,
+                                                           &sender_names, NULL,
+                                                           sender ? sender->policy->seclabel : bus->seclabel, receiver->policy->seclabel,
+                                                           message);
+                                r = log_commitf(bus->log, "Monitor :1.%llu is being disconnected as it does not have the resources to receive a message it subscribed to.", receiver->id);
+                                if (r)
+                                        return error_fold(r);
+                        } else {
                                 return error_fold(r);
+                        }
                 }
         }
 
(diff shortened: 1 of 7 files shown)


teg:
I updated with a few more commits.

src/util/user.h
```diff
@@ -24,6 +25,21 @@ enum {
...
+                return "FDs";
+        case USER_SLOT_MATCHES:
+                return "matches";
+        case USER_SLOT_OBJECTS:
+                return "objects";
+        default:
+                assert(0);
+        }
+}
+
```
dvdhrm: Why is that in the header?

src/util/user.c
```diff
@@ -312,6 +347,7 @@ static int user_compare(CRBTree *tree, void *k, CRBNode *rb) {
  * Return: 0 on success, negative error code on failure.
  */
 int user_registry_init(UserRegistry *registry,
+                       Log *log,
                        size_t n_slots,
                        const unsigned int *maxima) {
         static_assert(sizeof(*maxima) == sizeof(*registry->maxima),
```
dvdhrm: You sure you don't wanna assign `log` to `registry->log` here?

src/bus/driver.c
```diff
@@ -348,10 +348,22 @@ static int driver_monitor(Bus *bus, Peer *sender, Message *message) {
 
                 r = connection_queue(&receiver->connection, NULL, message);
                 if (r) {
-                        if (r == CONNECTION_E_QUOTA)
+                        if (r == CONNECTION_E_QUOTA) {
+                                NameSet sender_names = NAME_SET_INIT_FROM_OWNER(&sender->owned_names);
```
dvdhrm: `sender` can be NULL here, right?

src/bus/driver.c
```diff
@@ -348,10 +348,22 @@ static int driver_monitor(Bus *bus, Peer *sender, Message *message) {
...
+                                NameSet receiver_names = NAME_SET_INIT_FROM_OWNER(&receiver->owned_names);
+
                                 connection_shutdown(&receiver->connection);
-                        else
+
+                                log_append_here(bus->log, LOG_WARNING, 0);
+                                bus_log_append_transaction(bus, sender->id, receiver->id, &sender_names, &receiver_names,
+                                                           sender->policy->seclabel, receiver->policy->seclabel,
+                                                           message);
```
dvdhrm: Same here.

src/bus/driver.c
```diff
@@ -367,10 +379,21 @@ static int driver_send_unicast(Peer *receiver, Message *message) {
...
+                        NameSet receiver_names = NAME_SET_INIT_FROM_OWNER(&receiver->owned_names);
 
-                connection_shutdown(&receiver->connection);
+                        connection_shutdown(&receiver->connection);
+
+                        log_append_here(receiver->bus->log, LOG_WARNING, 0);
+                        bus_log_append_transaction(receiver->bus, ADDRESS_ID_INVALID, receiver->id, NULL, &receiver_names,
+                                                   receiver->bus->seclabel, receiver->policy->seclabel,
+                                                   message);
+                        r = log_commitf(receiver->bus->log, "Peer :1.%llu is being disconnected as it does not have the resources to receive a reply it requested.", receiver->id);
```
dvdhrm: This isn't necessarily a reply, is it? It can be a directed signal?

src/util/user.c
teg: Fixed.

src/bus/driver.c
teg: Indeed, and moreover, the receiver cannot own any names. Fixed.

src/bus/driver.c
teg: Fixed both.

src/util/user.h
teg: So far we kept the slot constants in the header with a TODO to move them elsewhere, so I kept this together.

teg:
I pushed an update addressing the comments.

src/util/user.c
```diff
@@ -208,6 +209,29 @@ static int user_ref_usage(User *user, UserUsage **usagep, User *actor) {
...
+
+        assert(slot < user->registry->n_slots);
+
+        log_appendf(log, "DBUS_BROKER_USER_CHARGE_USER=%u\n", user->uid);
+        log_appendf(log, "DBUS_BROKER_USER_CHARGE_ACTOR=%u\n", actor->uid);
+        log_appendf(log, "DBUS_BROKER_USER_CHARGE_AMOUNT=%u\n", amount);
+        log_appendf(log, "DBUS_BROKER_USER_CHARGE_N_ACTORS=%u\n", user->n_usages);
+        log_appendf(log, "DBUS_BROKER_USER_CHARGE_SLOT=%s\n", user_slot_to_string(slot));
+        log_appendf(log, "DBUS_BROKER_USER_CHARGE_REMAINING=%u\n", user->slots[slot].n);
+        log_appendf(log, "DBUS_BROKER_USER_CHARGE_CONSUMED=%u\n", charge->usage ? charge->usage->slots[slot] : 0);
```
dvdhrm: Didn't we document these in log.c? Can you append them there?

PR #143, opened 2018-06-07
match/iterator: replace iterators with getters for destinations

teg:
This simplifies the code, and keeps anything not match-related (policy decisions) in the caller. It marginally improves the worst case performance, but to avoid the O(n) worst-case performance, we would need to introduce Bloom filters, so optimizing this further does not seem worth the effort.


teg:
Rebased.

Issue #144, opened 2018-06-17
Problem parsing Exec= lines with quotes in dbus service files


jnsh:
While investigating why [certain gnome-shell extension didn't work when using dbus-broker](https://github.com/andyholmes/gnome-shell-extension-gsconnect/issues/77), I discovered an issue with dbus .service files, where the command in the `Exec=` line contains quotes.

I confirmed the issue using dbus services with the following `Exec=` lines, each of which should send a notification with **TEST** in the notification title when activated. I used `gdbus wait --session --activate <service>` for service activation.

These two work without issue:
```
Exec=/bin/notify-send TEST
```
```
Exec=/bin/env notify-send TEST
```

The following sends the notification, but the quotation marks follow to the notification, i.e. notification title becomes **"TEST"**, while it should still be **TEST** without quotations:
```
Exec=/bin/notify-send "TEST"
```

With this one, the command doesn't seem to execute at all. The behavior appears to be the same with any service, where the `Exec=` line contains a command with space somewhere between the quotation marks. This is also causing the issue with the gnome-shell extension mentioned above:
```
Exec=/bin/sh -c "notify-send TEST"
```

All of the above work as intended when dbus-broker is disabled.
Tested with dbus-broker v13.

teg:
Thanks for the report. I'll look into this.

teg:
I could reproduce, and the above patch fixes the issue for me. Thanks again for the report, must have been a pain to figure out!

dvdhrm:
Thanks a lot for the report! Feel free to open another report with some details on the issues you encounter. I will gladly try it myself and see what I can do!

PR #146, opened 2018-06-19
launch: properly unescape the Exec directive from .service files

teg:
The current logic was not correctly unquoating strings, as shown
by issue #144.

To quote the dbus sources, for the code that parses Exec

```
The results are defined to be the same as those you would get from
a UNIX98 /bin/sh, as long as the input contains none of the
unsupported shell expansions. If the input does contain such
expansions, they are passed through literally.
```

The same semantics is provided by a glib helper function, so we can
simply use that.

### DIFF
--- a/src/launch/main.c
+++ b/src/launch/main.c
@@ -52,8 +52,8 @@ struct Service {
         CRBNode rb_by_name;
         char *name;
         char *unit;
-        char **exec;
-        size_t n_exec;
+        size_t argc;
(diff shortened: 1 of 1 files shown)


dvdhrm:
We pass `argv` as `as` into `StartTransientUnit()`, right? So systemd does not interpret the information itself, anymore? At least that's how I read the transient-unit code. So looks all good to me. Thanks!

PR #147, opened 2018-06-20
Treat directed signals like method calls without replies

teg:
This only affects the case when the message transaction fails due to resource exhaustion. It used to be that if the receiver had a matching match installed they would be disconnected (treated as a solicited message). However, this semantics is not standardised, and installing matches for directed matches is not common practice.

Simplify the semantics by treating a directed signal just like a method call with the REPLY_NOT_EXPECTED flag set. This means that if the message could not be delivered it is dropped. This essentially implies that directed signals are recommended against if you want your client to be robust.


dvdhrm:
Yeah, as discussed previously, this seems like the only reasonable solution to this issue.

PR #150, opened 2018-06-22
policy: implement send_broadcast, min_fds and max_fds parameters

teg:
Internally, the implementation is symmetric, allowing recv_broadcast to
be supported too. However, that is not supported by the policy language,
nor the parser.

Also, extend the message logging to include the number of file descriptors,
as that can now influence policy decisions.

### DIFF
--- a/src/bus/policy.c
+++ b/src/bus/policy.c
@@ -10,6 +10,7 @@
 #include "bus/name.h"
 #include "bus/policy.h"
 #include "dbus/protocol.h"
+#include "util/common.h"
 #include "util/error.h"
 #include "util/selinux.h"
 
@@ -27,9 +28,12 @@ C_DEFINE_CLEANUP(PolicyXmit *, policy_xmit_free);
 
 static int policy_xmit_new(PolicyXmit **xmitp,
                            unsigned int type,
+                           unsigned int broadcast,
                            const char *path,
                            const char *interface,
-                           const char *member) {
+                           const char *member,
+                           uint64_t min_fds,
+                           uint64_t max_fds) {
         _c_cleanup_(policy_xmit_freep) PolicyXmit *xmit = NULL;
         size_t n_path, n_interface, n_member;
         void *p;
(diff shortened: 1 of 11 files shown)


src/launch/config.c
```diff
@@ -380,6 +380,8 @@ static int config_parser_attrs_allow_deny(ConfigState *state, ConfigNode *node,
         char *t;
         int r;
 
+        node->allow_deny.max_fds = UINT64_MAX;
+
```
dvdhrm: meh... That doesn't belong here... Can't we put that in config_node_new() just like config_node_free() does the cleanup?

src/bus/policy.c
```diff
@@ -13,6 +13,12 @@
 #include "util/error.h"
 #include "util/selinux.h"
 
+enum {
+        CONFIG_TRISTATE_UNSET,
+        CONFIG_TRISTATE_YES,
+        CONFIG_TRISTATE_NO,
```
dvdhrm: CONFIG? That should by `POLICY_TRISTATE_*` shouldn't it?

src/bus/policy.c
```diff
@@ -450,45 +469,57 @@ static int policy_registry_import_batch(PolicyRegistry *registry,
...
                             &verdict.priority,
                             &name_str,
                             &path,
                             &interface,
                             &member,
-                            &type);
+                            &type,
+                            &broadcast,
+                            &min_fds,
+                            &max_fds);
```
dvdhrm: Here...

src/bus/policy.c
```diff
@@ -450,45 +469,57 @@ static int policy_registry_import_batch(PolicyRegistry *registry,
...
```
dvdhrm: ...and here.

src/bus/policy.c
dvdhrm: So I dislike that we treat this as a common definition. We have this type-definition in the launcher, and now here as well, and we silently assume that the wire uses the same definition. Can we get some common solution here?

I marked the two places below where we read this type from the variant without any type-conversion (since both types match right now). Shouldn't there be some check that the value is not higher than 2?

I would be fine with placing this in src/util/common.h and naming it something like `UTIL_TRISTATE_*`.

src/launch/config.c
teg: Agreed. Fixed.

src/bus/policy.c
teg: Agreed. Changed.

PR #151, opened 2018-06-27
Gracefully handle ReloadConfig and SIGHUP when the config is invalid

teg:
Make sure we don't trigger a hard failure, but refuse to load invalid config. We may want to audit the config parser and consider more cases invalid rather than silently ignored, but that would be a follow-up.

```
Jun 27 17:32:44 teg-x270 systemd[1]: Reloading D-Bus System Message Bus.
Jun 27 17:32:44 teg-x270 dbus-broker-launch[763]: Invalid XML in /usr/share/dbus-1/system.conf +141: mismatched tag
Jun 27 17:32:44 teg-x270 busctl[1729]: Config invalid. Reload ignored.
Jun 27 17:32:44 teg-x270 systemd[1]: dbus-broker.service: Control process exited, code=exited status=1
Jun 27 17:33:57 teg-x270 dbus-broker-launch[763]: Caught SIGHUP
Jun 27 17:33:57 teg-x270 dbus-broker-launch[763]: Invalid XML in /usr/share/dbus-1/system.conf +141: mismatched tag
Jun 27 17:33:57 teg-x270 dbus-broker-launch[763]: Invalid configuration. SIGHUP ignored.
```

### DIFF
--- a/src/broker/controller-dbus.c
+++ b/src/broker/controller-dbus.c
@@ -528,6 +528,26 @@ static int controller_dispatch_reply(Controller *controller, uint32_t serial, co
         return 0;
 }
 
+static int controller_dispatch_error(Controller *controller, uint32_t serial, const char *error_name, Message *message) {
+        ControllerReload *reload;
+        int r;
+
+        reload = controller_find_reload(controller, serial);
+        if (!reload)
+                return CONTROLLER_E_UNEXPECTED_REPLY;
+
+        if (strcmp(error_name, "org.bus1.DBus.Controller.Error.InvalidConfig"))
+                return CONTROLLER_E_UNEXPECTED_ERROR;
+
+        r = controller_reload_invalid(reload);
+        if (r)
+                return error_trace(r);
+
+        controller_reload_free(reload);
+
+        return 0;
+}
+
 int controller_dbus_dispatch(Controller *controller, Message *message) {
         Connection *connection = &controller->connection;
         int r;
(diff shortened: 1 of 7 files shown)


src/broker/controller-dbus.c
```diff
@@ -528,6 +528,26 @@ static int controller_dispatch_reply(Controller *controller, uint32_t serial, co
...
+static int controller_dispatch_error(Controller *controller, uint32_t serial, const char *error_name, Message *message) {
+        ControllerReload *reload;
+        int r;
+
+        reload = controller_find_reload(controller, serial);
+        if (!reload)
+                return CONTROLLER_E_UNEXPECTED_REPLY;
+
+        if (strcmp(error_name, "org.bus1.DBus.Controller.Error.InvalidConfig"))
+                return CONTROLLER_E_UNEXPECTED_ERROR;
```
dvdhrm: Errr, shouldn't we invalidate the reload regardless of the error? I mean, we got a reply that matches the serial of our request, it seems counter-intuitive to ignore them?

dvdhrm: On the other hand, getting an unknown error from our own controller seems like an error we should catch. But if we ever want to open up the interface to external controllers, I think we should catch all errors here and treat them as failed reload, right?

Nothing critical, though.

teg: Hm, I'd say we should consider error codes API and not allow people to add random ones. But let's burn that bridge when we get to it.

PR #152, opened 2018-06-28
Include sys/wait.h for WEXITED definition

kraj:
Signed-off-by: Khem Raj <raj.khem@gmail.com>


dvdhrm:
`stdlib.h` of glibc pulls these in, but the explicit dependency looks correct! Thanks! Might affect uclibc/etc. builds.

PR #153, opened 2018-06-28
use getenv on non-glibc systems

kraj:
musl doesnt implement secure version

Signed-off-by: Khem Raj <raj.khem@gmail.com>


dvdhrm:
Lets not do that. There is no reason to use `secure_getenv()` here. Lets just switch to `getenv()`. We do not expect to be run as setuid, so `getenv()` should be fine.

Do you want to prepare a patch? If not, I can prepare one as well.

dvdhrm:
Thanks a lot! Looks good!

PR #154, opened 2018-06-29
launch: Fixes for config-reload

dvdhrm:
Assorted cleanups/fixes for the config reloader of the launcher.


dvdhrm:
Pushed with the second patch dropped. Thanks!

PR #157, opened 2018-07-05
Better handle invalid usernames in the config parser

teg:
It turns out to be more common than we'd like to have usernames in dbus config that are not defined on the system. Improve the logging and caching of these invalid entries.

### DIFF
--- a/src/launch/nss-cache.c
+++ b/src/launch/nss-cache.c
@@ -40,7 +40,7 @@ typedef struct NSSCacheNode {
 
 static int nss_cache_node_new(NSSCacheNode **nodep, const char *name, uint32_t id) {
         NSSCacheNode *node;
-        size_t n_name = strlen(name);
+        size_t n_name = name ? strlen(name) : 0;
 
         node = malloc(sizeof(*node) + n_name + 1);
         if (!node)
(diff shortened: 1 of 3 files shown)


src/launch/nss-cache.c
```diff
@@ -338,15 +353,36 @@ int nss_cache_get_uid(NSSCache *cache, uint32_t *uidp, uint32_t *gidp, const cha
...
+                                if (r)
+                                        return r;
+                        }
+
                         return NSS_CACHE_E_INVALID_NAME;
                 }
-
-                r = nss_cache_add_user(cache, pw);
-                if (r)
-                        return r;
```
dvdhrm: Why move this call up? Now the uid-lookup looks different than the gid-lookup.

src/launch/nss-cache.c
```diff
@@ -114,45 +114,50 @@ static int nss_cache_add(CRBTree *tree_by_name,
...
-         * known, or if they point to different entries, we will *NOT* cache
-         * the new data. This situation implies a UID/GID conflict and there is
-         * no sane way to handle this, other than bypassing the cache. Worst
-         * case, you will end up calling into NSS all the time.
+         * entry), it is linked by one and the other is invalid, or it is not
+         * linked at all. However, if only one of both is known and valid, or
+         * if they point to different entries, we will *NOT* cache the new data.
+         * This situation implies a UID/GID conflict and there is no sane way to
+         * handle this, other than bypassing the cache. Worst case, you will
+         * end up calling into NSS all the time.
```
dvdhrm: This reads horribly. "Either the entry [...] is linked by one and the other is invalid [...]. However, if only one of both is known and valid, ...". The second part refers to the two different entries, the first part to the two different indices. But that is really hard to grasp.

src/launch/nss-cache.c
```diff
@@ -114,45 +114,50 @@ static int nss_cache_add(CRBTree *tree_by_name,
...
          */
-        if (node_by_name != node_by_id) {
+        if ((node_by_id && node_by_id->name && !node_by_name) ||
```
dvdhrm: This is wrong. `name` is embedded, it cannot be NULL. You *MUST* check for `node->pw.pw_name`, but you can't since you don't know whether this is a user or group. Why not rather check `c_rbnode_is_linked(&node->rb_by_name)`?

src/launch/nss-cache.c
```diff
@@ -114,45 +114,50 @@ static int nss_cache_add(CRBTree *tree_by_name,
...
+            (node_by_name && node_by_name->id != -1 && !node_by_id) ||
+            (node_by_name && node_by_id && node_by_name != node_by_id)) {
                 *nodep = NULL;
                 return 0;
         }
 
-        if (!node_by_name && !node_by_id) {
+        if (!node_by_name && !node_by_id && (name || id != -1)) {
```
dvdhrm: This last check is redundant, isn't it? Also, if neither `name` nor `id` is set, this function will not complain but return `NULL` and success. That is weird, isn't it?

dvdhrm:
Also lots of `comparing unsigned with signed integer` warnings in `nss-cache.c`. You cannot do `id != -1`, you must use `id != (uint32_t)-1` to silence those.

src/launch/nss-cache.c
teg: Rephrased.

src/launch/nss-cache.c
teg: Fixed.

src/launch/nss-cache.c
teg: The check is to deal with the case of passing in invalid both `name` and `id` (which would otherwise be added to the cache). I moved that to an explicit check at the top, and made it a failure.

teg:
Rebased, all comments should now be addressed.

teg:
Force pushed with a fixup for groups to work the same as users that had been leftout.

PR #160, opened 2018-07-10
broker: use parent seclabel for the driver

dvdhrm:
This makes the broker use the parent seclabel for the driver, as discussed offchannel. The reasoning is in the commit-message / inline-comment.

### DIFF
--- a/src/broker/broker.c
+++ b/src/broker/broker.c
@@ -84,7 +84,7 @@ int broker_new(Broker **brokerp, const char *machine_id, int log_fd, int control
         if (r)
                 return error_fold(r);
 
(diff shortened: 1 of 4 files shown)


src/broker/broker.c
```diff
@@ -84,7 +84,22 @@ int broker_new(Broker **brokerp, const char *machine_id, int log_fd, int control
         if (r)
                 return error_fold(r);
 
-        r = proc_get_seclabel(PROC_SELF, &broker->bus.seclabel, &broker->bus.n_seclabel);
+        /*
+         * XXX: We need the seclabel to run the broker as for 2 reasons: First,
```
dnicolodi: Typo: `run the broker as for 2 reasons` → `run the broker for two reasons`.

PR #166, opened 2018-08-27
launch: add back --version command line switch

Keruspe:
For backward compatibility (because of at-spi2)


teg:
This makes sense, the switch should not have been removed. Thanks!

PR #168, opened 2018-10-08
use the configured resource limits

teg:
It is the distro/admin's responsibility to ensure that these are reasonable.

### DIFF
--- a/src/launch/config.c
+++ b/src/launch/config.c
@@ -154,9 +154,6 @@ ConfigNode *config_node_free(ConfigNode *node) {
         case CONFIG_NODE_SERVICEDIR:
                 free(node->servicedir.path);
                 break;
-        case CONFIG_NODE_LIMIT:
-                free(node->limit.name);
-                break;
         case CONFIG_NODE_ALLOW:
         case CONFIG_NODE_DENY:
                 free(node->allow_deny.send_interface);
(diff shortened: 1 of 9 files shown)


src/launch/config.c
```diff
@@ -1109,12 +1136,26 @@ static void config_parser_end_fn(void *userdata, const XML_Char *name) {
                 break;
         }
 
+        case CONFIG_NODE_LIMIT: {
+                unsigned long long limit;
+                char *end;
+
+                errno = 0;
+                limit = strtoull(state->current->cdata, &end, 10);
+                if (end != state->current->cdata && *end == '\0' && errno == 0)
```
dvdhrm: Errrr, this hard-codes `sizeof(long long)` in user-visible stuff. This is 64bit everywhere, I know, so probably fine. But... eh... if "long long" is not 64bit, this suddenly has different maximums depending on the machine. Not sure that's what we want.

Can we make this a util/ helper? util_strtou64()? and then "static_assert(...)"?

teg: Will do.

teg:
Force-pushed. Fixes the saturating multiplication as well as uses 64-bit integers everywhere, rather than unsigned long long.

dvdhrm:
Looks good to me! Thanks!

PR #170, opened 2018-10-25
launcher: replace g_shell with c-shquote

teg:
c-shquote is a stand-alone library, compatible with glib's g_shell
helpers. Use that to parse the argv's to pass to PID1.

This only leaves he ini parser before we could potentially drop the
glib dependency.


src/launch/main.c
```diff
@@ -168,17 +167,8 @@ static int service_update(Service *service, const char *unit, size_t argc, char
         }
 
         if (argc > 0) {
-                service->argv = calloc(1, argc * sizeof(char*));
-                if (!service->argv)
-                        return error_origin(-ENOMEM);
-
+                service->argv = argv;
```
dvdhrm: We usually don't `consume` parameters like this, do we? We sometimes do things like `foobar_new_for_sth()` or `_with_()` which consumes the main argument. However, we never do this with long argument lists like here. Any reason not to dup the value? Or is that fragile, because it has pointers inside of it? Alternative would be to pass a double pointer and clear the pointer when we consume it. Not sure, maybe this code is fine. Comments?

teg: I think the only reasonable alternative would be to use the same logic as before and dup each individual argument. I think I'd be fine with that, and I agree that this is a bit too unidiomatic. I'll rework it.

teg: Force-pushed.

PR #178, opened 2018-12-03
Be stricter about FD passing

teg:
Only send/recevie FDs from peers who have negotiated support for that, and consider receipt of FDs during authentication a protocol violation.


src/dbus/connection.c
```diff
@@ -259,6 +260,18 @@ int connection_dequeue(Connection *connection, Message **messagep) {
                 return error_fold(r);
         }
 
+        if (fdlist_count(message->fds) > 0) {
```
dvdhrm: `message` might be NULL. This shouldn't pass the test suite.

teg: Indeed. Fixed and force-pushed. Still issues outstanding.

teg:
Force pushed. The bug fixes are also available in #179.

PR #180, opened 2018-12-05
More disconnect logging

teg:
Be more fine-grained about the protocol violations causing forcible disconnect.

This is on top of #178.


teg:
Updated to keep driver_end_read() in driver.c. We still rely on messages being verified before being passed to the driver, but let's keep the helper around for documentation purposes.

dvdhrm:
I cherry-picked the patches, as they no longer applied cleanly (and are mixed with other PRs). I also pushed 2 style-fixes on top. Lemme know if you have any comments on them.

Otherwise, looks all good! We lack some error-handling in the controller, but we do properly fold everything, so it's just not as fine-grained as the peer-error-handling. We should fix that some day..

Thanks!

PR #182, opened 2018-12-31
Minor name correction

Jacalz:
I think this was my fault for signing off with my username and not my real
name. Did not think of that at the time, anyhow this is corrected now :)

Happy New Years to you all! :fireworks:


Jacalz:
cc @dvdhrm

Issue #183, opened 2019-01-11
[question] Run dbus-broker without systemd running


jkonecny12:
Hello,

I'm working as the Anaconda (Fedora OS installer) [0] developer. Recently we are migrating Anaconda to a modular system where modules (parts of the installer) will connect to each other and communicate by dbus, this is already implement for most of the modules. We are using dbus [1] and we want to migrate to dbus-broker as other applications in Fedora. In fact we have a beta blocker bug for Fedora 30 which would be solved by this.

The problem is that one of our use-cases is building installation images. Not only, but we also support building these images in a mock [2] environment which is basically container so it doesn't have systemd running.

Could you please tell me if there is a way how to start our own dbus session without involving systemd at all? If it is, could you give me some pointers where to start?

[0]: https://github.com/rhinstaller/anaconda
[1]: https://github.com/rhinstaller/anaconda/blob/master/pyanaconda/dbus/launcher.py#L107
[2]: https://github.com/rpm-software-management/mock/wiki

dvdhrm:
Hey!

The core of dbus-broker (the `dbus-broker` binary) has no external dependencies at all, other than a linux kernel. It is a complete bus implementation according to the dbus specification (see `dbus-broker(1)` man-page).

However, most of the time you expect an environment around a bus that provides many more features atop of the pure bus broker. This is why we also provide `dbus-broker-launch` (see `dbus-broker-launch(1)` man-page). This is a compatibility interface to dbus-daemon, that tries to provide an environment as close to what dbus-daemon provides.
When writing this compatibility launcher, we decided to strip all legacy features and instead built on top of what is present on a modern Fedora system. This means, this implementation requires an activation environment which includes a running systemd manager.

If all you need is a simple bus without any features on top, you can easily run `dbus-broker` directly without any launcher. However, if you need dbus-activation, then you need a bus controller that spawns dbus-services when they are activated. The compatibility launcher uses systemd to do this. If you want to spawn dbus-services without systemd, you need a suitable replacement. I am not aware of any replacement on Fedora, but you can of course roll your own. In this case it is up to you to implement the controller interface documented in `dbus-broker(1)`, spawn the dbus-services, track and monitor them.

What exactly are your requirements? Do you need dbus-activation? Why is there no systemd available? What do you use to spawn the different modules running in anaconda?

Cheers!
David

teg:
[...]

PR #186, opened 2019-01-14
[RFC] launch: allow running without a scope

teg:
The scope (system or user) determines which systemd instance to
connect to (the system or user instance) in order to start
dbus-activated services.

If no scope is specified we were defaulting to "system".

Change the logic to allow running without a scope at all and make
that the new default (all users specify one, as far as I know).

When running without a scope, name activation is disabled. That
means that we no longer expose names as activatable (but warn
from the launcher in case such names are conigured), and we ignore
(and warn about) calls to change the activation environment.

Apart from just making sense to me, this may be relevant to issue #183.

### DIFF
--- a/src/launch/main.c
+++ b/src/launch/main.c
@@ -49,6 +49,12 @@ enum {
         MANAGER_E_INVALID_SERVICE_FILE,
 };
 
+enum {
+        MANAGER_SCOPE_NONE,
+        MANAGER_SCOPE_USER,
+        MANAGER_SCOPE_SYSTEM,
+};
+
 typedef enum {
         SERVICE_STATE_PENDING,
         SERVICE_STATE_CURRENT,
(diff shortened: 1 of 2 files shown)


jkonecny12:
Nice idea!

I would also add the `--controller` from `dbus-broker`. It feels logical to me that if you start with `--scope=none` to avoid systemd dependency than you also don't want to use their socket unit and having to implement their socket activation for every such project seems unnecessary. 
What do you think?

src/launch/main.c
```diff
@@ -801,6 +808,11 @@ static int manager_on_set_activation_environment(Manager *manager, sd_bus_messag
         _c_cleanup_(sd_bus_message_unrefp) sd_bus_message *method_call = NULL;
         int r;
 
+        if (main_arg_scope == MANAGER_SCOPE_NONE) {
+                fprintf(stderr, "Ignoring request to set activation environment outside of scope.\n");
+                return 0;
```
dvdhrm: Errr, this should not print a message. We should just pretend to honor the call and silently return.

teg:
> I would also add the `--controller` from `dbus-broker`. It feels logical to me that if you start with `--scope=none` to avoid systemd dependency than you also don't want to use their socket unit and having to implement their socket activation for every such project seems unnecessary.
> What do you think?

I created a separate PR (#188) to discuss this, let's move that part of the discussion over there. To be clear: I'm open to the idea, but not entirely convinced we should add the extra code.

PR #188, opened 2019-01-17
[RFC] launch: allow listening socket to be given on the commandline

teg:
This follows the style of the broker API, and makes it slightly more convenient to spawn the launcher manually without systemd activation.

This might not be worth the extra codepaths, as the systemd socket passing API is sort of trivial, and this still requires the caller to actually create the socket etc manually.

### DIFF
--- a/docs/dbus-broker-launch.rst
+++ b/docs/dbus-broker-launch.rst
@@ -21,7 +21,7 @@ DESCRIPTION
 ===========
 
 **dbus-broker-launch** is a launcher for **dbus-broker**, spawning and managing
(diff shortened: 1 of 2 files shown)


jkonecny12:
Thanks a lot for this change. It may make our life easier. 
However, before this will be merged we should discuss this in person on DevConf to find out if we want to use this or we want to use `dbus-broker` instead.
And if you in general want to support this.

dvdhrm:
@teg Can we close this?

jkonecny12:
We won't be using this solution after all. So you don't need to support this only for us.

PR #189, opened 2019-01-22
Treat unreadable service files as invalid

yaneti:
This is a followup to a0869bdb8bb6dde92374bc12f5428913d5357653.
If we return 0 the parsing is considered complete and bad thing happen in manager_load_service_file


yaneti:
Please feel free not to merge the fix and deal with all the cases as you see fit.

yaneti:
I've mangled this simple patch enough I think. :) added a sign-off .  If its still not up to par please apply whatever you deem necessary to avoid the crashing :)

dvdhrm:
This is now merged. Thanks a lot!

PR #195, opened 2019-03-28
launch: improve the logging

teg:
Suppress some redundant messages and use the journal more.


teg:
> Why are you ignoring the return code of `log_commitf()` almost everywhere?

Oversight. Fixed and force-pushed.

dvdhrm:
Adjusted minor bits and then merged. Thanks a lot!

PR #213, opened 2019-10-25
Fix running dbus-broker with <selinux> in the policy

### LINKED ISSUE #212: dbus-broker-launch asserts if policy contains <selinux> node
dbus-broker-launch (both v19 and latest master) fails with:

> Oct 22 12:02:51 localhost dbus-broker-launch[885]: dbus-broker-launch: ../dbus-broker-19/src/launch/policy.c:232: policy_import_verdict: Assertion `cnode->parent->type == CONFIG_NODE_POLICY' failed.

Minimal config file to reproduce:

```
<!DOCTYPE busconfig PUBLIC "-//freedesktop//DTD D-BUS Bus Configuration 1.0//EN"
 "http://www.freedesktop.org/standards/dbus/1.0/busconfig.dtd">
<busconfig>
  <selinux>
    <associate own="com.example1" context="system_u:object_r:example1_t" />
    <associate own="com.example2" context="system_u:object_r:example2_t" />
  </selinux>
</busconfig>
```

It seems to me that it is at the same time asserting that the parent of the cnode being currently parsed is both CONFIG_NODE_SELINUX and a CONFIG_NODE_POLICY:

https://github.com/bus1/dbus-broker/blob/master/src/launch/policy.c#L603

```
c_assert(cnode->parent->type == CONFIG_NODE_SELINUX);
...
policy_import_verdict(policy, record, cnode);
```

https://github.com/bus1/dbus-broker/blob/master/src/launch/policy.c#L232
```
c_assert(cnode->parent->type == CONFIG_NODE_POLICY);
```

Unless I am missing something?

bluca:
dbus-broker crashes on startup if the policy includes ```<selinux>``` elements. I have verified on a yocto-based distro that with the following fixes everything works. A process with the correct label is allowed to own an interface, and a process without the correct label is not.

Fixes #212

### DIFF
--- a/src/launch/policy.c
+++ b/src/launch/policy.c
@@ -1138,7 +1138,7 @@ int policy_export(Policy *policy, sd_bus_message *m, uint32_t *at_console_uids, 
                 r = sd_bus_message_append(m, "(ss)",
                                           i_record->selinux.name,
                                           i_record->selinux.context);
-                if (r)
+                if (r < 0)
                         return error_origin(r);
         }
 
--- a/src/launch/policy.c
+++ b/src/launch/policy.c
@@ -613,8 +613,6 @@ static int policy_import_selinux(Policy *policy, ConfigNode *cnode) {
         if (r)
                 return error_trace(r);
 
-        policy_import_verdict(policy, record, cnode);
-
         record->selinux.name = cnode->associate.own;
         record->selinux.context = cnode->associate.context;
 


bluca:
CI error is from one of the submodules:

```
../../src/subprojects/c-dvar/src/c-dvar-writer.c: In function 'c_dvar_try_vwrite':
../../src/subprojects/c-dvar/src/c-dvar-writer.c:207:25: error: ISO C90 forbids mixed declarations and code [-Werror=declaration-after-statement]
  207 |                         static_assert(sizeof(double) == sizeof(uint64_t),
```

dvdhrm:
Whoops, this is embarrassing. Does nobody in Fedora actually use the selinux policies or do we fail to load them? @teg do you know?

Anyway, both fixes look correct to me. Thanks a lot!

bluca:
No problem! I'm told it's pretty rare to see selinux elements in policies. One other reason might be that I think those asserts are compiled out if NDEBUG is used - not sure if Fedora builds with that though.

dvdhrm:
Good point, I think the release builds set it, which might explain that we did not hit it.

bluca:
@teg gentle ping. Any chance this could be looked at? Thanks!

teg:
Sorry for dropping the ball on this! Looking good to me

bluca:
No problem, thanks for having a look!

bluca:
rebased, so that the CI run can go through thanks to the fixes in master.

dvdhrm:
Looks good, thanks a lot!

PR #217, opened 2019-12-20
launch/service: support noop activation files

teg:
If SystemdService and Exec are both missing from the service files
then we still treat the name as activatable, but no explicit action
is needed from us in order to start it (it is going to be started
by the system, and activation is only used to avoid a race).

See discussion here: https://github.com/systemd/systemd/pull/14405


dvdhrm:
Love it! Reduced one more special case and instead assigns a natural behavior to self-activating service files that carry no activation information.

I restarted the CI and will merge it if it passes.

dvdhrm:
Meh, CI probably needs a rebase on master.

PR #238, opened 2020-08-27
util/selinux: Change auditing based on the log type.

pebenito:
The SELinux log callback includes a message type.  Not all messages are
auditable and those that are don't have the different audit types.  Update
the auditing accordingly.

If the message is not auditable, write it to stderr.

Signed-off-by: Chris PeBenito <chpebeni@linux.microsoft.com>

### DIFF
--- a/src/util/audit.c
+++ b/src/util/audit.c
@@ -92,6 +92,7 @@ int util_audit_drop_permissions(uint32_t uid, uint32_t gid) {
 
 /**
  * util_audit_log() - log a message to the audit subsystem
+ * @type:       the audit message type, see enum in audit.h
  * @message:    the message to be logged
  * @uid:        the UID of the user causing the message to be logged
  *
@@ -100,11 +101,24 @@ int util_audit_drop_permissions(uint32_t uid, uint32_t gid) {
  *
  * Return: 0 on success, or a negative error code on failure.
  */
-int util_audit_log(const char *message, uid_t uid) {
-        int r;
+int util_audit_log(int type, const char *message, uid_t uid) {
+        int r, audit_type;
 
-        if (audit_fd >= 0) {
-                r = audit_log_user_avc_message(audit_fd, AUDIT_USER_AVC, message, NULL, NULL, NULL, uid);
+        switch(type) {
+        case UTIL_AUDIT_TYPE_AVC:
+                audit_type = AUDIT_USER_AVC;
+                break;
+        case UTIL_AUDIT_TYPE_SELINUX_ERROR:
+                audit_type = AUDIT_USER_SELINUX_ERR;
+                break;
+        case UTIL_AUDIT_TYPE_NOAUDIT:
+        default:
+                audit_type = 0;
+                break;
+        }
+
+        if (audit_fd >= 0 && type != UTIL_AUDIT_TYPE_NOAUDIT) {
+                r = audit_log_user_avc_message(audit_fd, audit_type, message, NULL, NULL, NULL, uid);
                 if (r <= 0)
                         return error_origin(-errno);
         } else {
--- a/src/util/audit.h
+++ b/src/util/audit.h
@@ -7,8 +7,14 @@
 #include <c-stdaux.h>
 #include <stdlib.h>
 
+enum {
+        UTIL_AUDIT_TYPE_NOAUDIT,
+        UTIL_AUDIT_TYPE_AVC,
+        UTIL_AUDIT_TYPE_SELINUX_ERROR,
+};
+
 int util_audit_drop_permissions(uint32_t uid, uint32_t gid);
-int util_audit_log(const char *message, uid_t uid);
+int util_audit_log(int type, const char *message, uid_t uid);
 
 int util_audit_init_global(void);
 void util_audit_deinit_global(void);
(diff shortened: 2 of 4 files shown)


src/util/audit.h
```diff
@@ -7,8 +7,14 @@
 #include <c-stdaux.h>
 #include <stdlib.h>
 
+enum {
+        AUDIT_TYPE_NOAUDIT,
+        AUDIT_TYPE_AVC,
+        AUDIT_TYPE_SELINUX_ERROR,
```
dvdhrm, line 13: This should probably use the `UTIL_AUDIT_*` prefix to adhere to the strict namespacing.

src/util/audit.c
```diff
@@ -100,11 +101,23 @@ int util_audit_drop_permissions(uint32_t uid, uint32_t gid) {
...
+        switch(type) {
+                case AUDIT_TYPE_AVC:
+                        audit_type = AUDIT_USER_AVC;
+                        break;
+                case AUDIT_TYPE_SELINUX_ERROR:
+                        audit_type = AUDIT_USER_SELINUX_ERR;
+                        break;
+                case AUDIT_TYPE_NOAUDIT:
+                default:
+                        audit_type = 0;
```
dvdhrm, line 116: We don't indent the switch-cases, so this can be reduced by 8ch. And we usually break the default-clauses as well.

src/util/audit.c
```diff
@@ -100,11 +101,23 @@ int util_audit_drop_permissions(uint32_t uid, uint32_t gid) {
...
+        }
 
-        if (audit_fd >= 0) {
-                r = audit_log_user_avc_message(audit_fd, AUDIT_USER_AVC, message, NULL, NULL, NULL, uid);
+        if (audit_fd >= 0 && audit_type > 0) {
```
dvdhrm, line 119: Maybe be more explicit and use `type != AUDIT_TYPE_NOAUDIT`? I am a bit unsure about relying on external enums to not be 0 (especially since I did not find an explicit statement in libaudit that guarantees such ABI).

src/util/selinux.c
```diff
@@ -288,18 +288,30 @@ int bus_selinux_check_send(BusSELinuxRegistry *registry,
...
+        switch(type) {
+                case SELINUX_AVC:
+                        audit_type = AUDIT_TYPE_AVC;
+                        break;
+                case SELINUX_ERROR:
+                        audit_type = AUDIT_TYPE_SELINUX_ERROR;
+                        break;
+                default:
+                        /* not an auditable message. */
+                        audit_type = AUDIT_TYPE_NOAUDIT;
```
dvdhrm, line 308: Same style comments as above.

dvdhrm (approved):
Thanks a lot for the PR! Looks good to me. Minor style comments inline, but I will gladly fix them myself in a follow-up commit.

dvdhrm:
Before I merge this, can you give me a short background info on this? Is there any harm in passing things to libaudit which do not match AVC/ERROR? Or is this just about matching the right error to the right audit-call, so we don't accidentally trigger audit messages for things that should not have triggered them?

pebenito:
Audit messages are security events, so informational messages generally aren't audited. Not everything coming out of the libselinux log callback should be audited. This is a problem for systems that scan audit logs for certain audit types to initiate incident actions.

When you load a new SELinux policy you get this:
```
audit[364]: USER_AVC pid=364 uid=999 auid=4294967295 ses=4294967295 subj=system_u:system_r:system_dbusd_t msg='avc:  received policyload notice (seqno=2) exe="/usr/bin/dbus-broker" sauid=999 hostname=? addr=? terminal=?'
```
USER_AVC is not the correct audit message, since nothing is being denied. It's the code acknowledging that it has done operations in response to a policy load.

This PR is first step to comprehensively fix this problem. It does have a residual problem due to a libselinux issue. libselinux isn't providing its own log message types so callers can determine if something should be audited. In the above case, it really should be audited as USER_MAC_POLICY_LOAD, but you can't tell in the libselinux callback because it currently has the policy load message as an "INFO" log level. I have a [patch](https://lore.kernel.org/selinux/20200827125839.79006-1-chpebeni@linux.microsoft.com/T/#u) to that upstream to fix it. I also need to do a PR to libaudit USER_MAC_STATUS audit event type to audit one more event coming from libselinux.

Once libselinux and libaudit is fixed I'll follow up with another PR to add on additional audit message types.

src/util/audit.c
pebenito: That would work if something comes in as `AUDIT_TYPE_NOAUDIT`, but if a type isn't an enum value, then this wouldn't work. Maybe set `audit_type` to `-1` for AUDIT_TYPE_NOAUDIT/default?

dvdhrm:
> Audit messages are for security events, so informational messages generally aren't audited. Not everything coming out of the libselinux log callback should be audited. This is a problem for systems that scan audit logs for certain audit types to initiate incident actions.

Yeah, fair enough. We should definitely match the logging types. Thanks for the background info!

src/util/audit.c
dvdhrm: Why would anyone call this with an invalid enum value? This is our internal API, and we control all callers. They only use values we define, right?

pebenito: My mode is always to have defensive programming. If you feel this is overkill I get it in this case.

pebenito:
I think this addresses all of your comments.

dvdhrm (approved):

dvdhrm:
Thank you very much! Not sure when I will do a next release, but I guess sometime September.

Issue #245, opened 2020-11-17
Abort when killing dbus service and monitoring bus


benzea:
I was doing some bus monitoring using dbus-monitor with various watches. I am not sure I can still recover the exact watch expression and I changed it around quite a bit before. In the end I hit Ctrl+C on the service I was debugging (not dbus-monitor), and dbus-broker died with the following.

```
Nov 17 11:27:24 ben-x1c7 dbus-broker-launch[9968]: dbus-broker: ../src/bus/match.c:1016: match_registry_deinit: Assertion `c_assert_result && "c_rbtree_is_empty(&registry->monitor_tree)"' failed.
Nov 17 11:27:25 ben-x1c7 dbus-broker-launch[9950]: ERROR launcher_run @ ../src/launch/launcher.c +1407: Return code 1
Nov 17 11:27:25 ben-x1c7 dbus-broker-launch[9950]:       run @ ../src/launch/main.c +152
Nov 17 11:27:25 ben-x1c7 dbus-broker-launch[9950]:       main @ ../src/launch/main.c +178
Nov 17 11:27:25 ben-x1c7 dbus-broker-launch[9950]: Exiting due to fatal error: -131
```

The backtrace is pretty boring to be honest.

```
Thread 1 (Thread 0x7f4acd959e40 (LWP 9968)):
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007f4acda44895 in __GI_abort () at abort.c:79
#2  0x00007f4acda44769 in __assert_fail_base (fmt=0x7f4acdbb20b8 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", assertion=0x55b5393cc258 "c_assert_result && \"c_rbtree_is_empty(&registry->monitor_tree)\"", file=0x55b5393cbe86 "../src/bus/match.c", line=1016, function=<optimized out>) at assert.c:92
#3  0x00007f4acda53e76 in __GI___assert_fail (assertion=assertion@entry=0x55b5393cc258 "c_assert_result && \"c_rbtree_is_empty(&registry->monitor_tree)\"", file=file@entry=0x55b5393cbe86 "../src/bus/match.c", line=line@entry=1016, function=function@entry=0x55b5393cc350 <__PRETTY_FUNCTION__.5> "match_registry_deinit") at assert.c:101
[...]

PR #246, opened 2020-11-18
bus/match: flush monitor matches on disconnect

dvdhrm:
(CC: @benzea @teg #245)

We did not correctly flush monitor matches on peer disconnect, but only the normal subscription objects. Fix this and properly flush all matches in a registry on disconnect.

See the commit message for background information.

I have not tested this, yet. But I think this is quite likely the bug @benzea saw. Will report back when I tested it.

### DIFF
--- a/src/bus/match.c
+++ b/src/bus/match.c
@@ -1137,16 +1137,35 @@ static void match_registry_by_path_flush(MatchRegistryByPath *registry) {
 }
 
 /**
- * mach_registry_flush() - XXX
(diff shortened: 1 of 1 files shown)


teg (approved):
This looks clearly right to me. Sorry for introducing the bug way back when. Thanks for the fix!

dvdhrm:
Indeed, I verified it now. This does fix the issue and with this applied I cannot reproduce it anymore. I will also merge the audit-3.0 changes once they released their new version and then prep a dbus-broker release next week.

PR #251, opened 2021-01-19
[PENDING] util/selinux: Add additional audit message types.

dvdhrm:
I picked up the patch from @pebenito so we don't forget about it. The original PR was #240 and was closed automatically when we switched the main branch.

I also updated the `meson.build` dependency to the now released `audit-3.0`. Once that hits the bigger distro channels, I think we can merge this. I would like to wait for the `libselinux` release as well, but this seems like something we can carry with an `#ifndef` until they prep a new release.

Anyway, leaving this here so we don't forget about this.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -46,7 +46,7 @@ dep_thread = dependency('threads')
 
 use_audit = get_option('audit')
 if use_audit
-        dep_libaudit = dependency('audit', version: '>=2.7')
+        dep_libaudit = dependency('audit', version: '>=3.0')
         dep_libcapng = dependency('libcap-ng', version: '>=0.6')
 endif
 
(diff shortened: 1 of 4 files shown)


dvdhrm:
FYI, the `libselinux` release we need is `>3.1` (so probably `3.2`).

pebenito:
libselinux 3.2 is correct, in fact rc1 just came out: https://github.com/SELinuxProject/selinux/releases/tag/3.2-rc1

dvdhrm:
I updated this to use the 3.2 selinux release. The required changes are already in Fedora-34 & rawhide. My plan is to do one more release without this, then merge this and push it into F34 and rawhide.

dvdhrm:
Now merged. Thanks for the patches! This will be part of the next release, and thus make it into F34 and F35.

Issue #269, opened 2021-06-25
bus activated services that fail can just get `Could not activate remote peer` errors


cgwalters:
See https://github.com/coreos/rpm-ostree/pull/2932

If I explicitly force rpm-ostreed to fail during startup before claiming the bus name, clients can just see errors like:

```
[root@cosa-devsh ~]# rpm-ostree status
error: Could not activate remote peer.
```

But, if a service has an associated systemd unit (and they all do now right?) then dbus-broker could at least give us the exit status from systemd, the `StatusText` if present.  And if there's no `StatusText`, maybe the last line from the journal from the unit?

teg:
It would make sense to forward more information in this error message, see https://github.com/bus1/dbus-broker/blob/28af5ac996c94922fc49660e0efa955f57575187/src/launch/service.c#L202

There are three error cases:
 * `ServiceStart{,Transient}Unit` can fail (unit file maked/disabled/...)
 * We receive `JobRemoved` signal indicating the unit was not spawned successfully (exec() failed,...)
 * The unit enters the `failed` `ActiveState` (probably your case, where the unit fails after being started, but before grabbing the DBus name).

systemd passes us different metadata in each case, so not sure we can always pass on anything helpful, but makes sense to go pass on whatever useful information we receive (not sure about querying more stuff from systemd/journald though).

cgwalters:
> The unit enters the failed ActiveState (probably your case, where the unit fails after being started, but before grabbing the DBus name).

Yep, this case.

I don't have a strong opinon on this.  If it's hard to fix, feel free to ignore/close.  I'm going to work on that linked PR to skip DBus for starting the daemon.

dvdhrm:
> > The unit enters the failed ActiveState (probably your case, where the unit fails after being started, but before grabbing the DBus name).
> 
> Yep, this case.
> 
> I don't have a strong opinon on this. If it's hard to fix, feel free to ignore/close. I'm going to work on that linked PR to skip DBus for starting the daemon.

The biggest issue is that we don't want to perform IPC from within the IPC implementation. This easily leads to deadlocks. In this case here, it is less critical, because it would be an asynchronous call in the launcher, and thus not block the broker itself. However, it can still deadlock, if the call out triggers a dbus activation on the systemd side, which then calls back and possibly waits on that stalled call. All quite unlikely, but we have run into many such problems with NSS and user-uid resolution, and hence we prefer to avoid performing any IPC from within the broker.

[...]

PR #277, opened 2021-12-03
launcher: propagate readable error message on service activation failure

bluca:
Example:

1) ConditionPathExists=/tmp/marker fails (note: with change in systemd to return `skipped`)
2) AssertPathExists=/tmp/marker fails
3) service immediately exits with failure before taking ownership of bus

```
root@image:~# busctl call com.ClientServerExample /com/ClientServerExample com.ClientServerExample Hello
Call failed: Could not activate remote peer: job result: skipped.
root@image:~# nano /etc/systemd/system/hello.service 
root@image:~# systemctl daemon-reload
root@image:~# busctl call com.ClientServerExample /com/ClientServerExample com.ClientServerExample Hello
Call failed: Could not activate remote peer: job result: assert.
root@image:~# nano /etc/systemd/system/hello.service 
root@image:~# systemctl daemon-reload
root@image:~# busctl call com.ClientServerExample /com/ClientServerExample com.ClientServerExample Hello
Call failed: Could not activate remote peer: job result: failed.
```

Does this sound sensible?

Fixes https://github.com/bus1/dbus-broker/issues/269

### DIFF
--- a/src/launch/service.c
+++ b/src/launch/service.c
@@ -17,6 +17,7 @@
 #include <c-stdaux.h>
 #include <stdlib.h>
 #include <systemd/sd-bus.h>
+#include "broker/controller.h"
 #include "catalog/catalog-ids.h"
 #include "launch/service.h"
 #include "util/error.h"
@@ -188,7 +189,7 @@ static void service_discard_activation(Service *service) {
         service->slot_watch_jobs = sd_bus_slot_unref(service->slot_watch_jobs);
 }
 
-static int service_reset_activation(Service *service) {
+static int service_reset_activation(Service *service, int bus1_error) {
         _c_cleanup_(c_freep) char *object_path = NULL;
         int r;
 
@@ -198,13 +199,6 @@ static int service_reset_activation(Service *service) {
         if (r < 0)
                 return error_origin(-errno);
 
-        /*
-         * XXX: Ideally, we would include more detailed error information as
-         *      payload. This would allow the broker to forward this in the
-         *      dbus error to each pending transaction.
-         *      So far, we did not define such errors, so this is left for the
-         *      future.
-         */
         r = sd_bus_call_method(service->launcher->bus_controller,
                                NULL,
                                object_path,
@@ -212,8 +206,9 @@ static int service_reset_activation(Service *service) {
                                "Reset",
                                NULL,
                                NULL,
-                               "t",
-                               service->last_serial);
+                               "ts",
+                               service->last_serial,
+                               bus1_error_to_string(bus1_error));
         if (r < 0)
                 return error_origin(r);
 
@@ -260,7 +255,7 @@ static int service_watch_jobs_handler(sd_bus_message *message, void *userdata, s
                  * Our job failed. Forward this information to the broker so it
                  * can fail pending activations.
                  */
-                r = service_reset_activation(service);
+                r = service_reset_activation(service, !strcmp(result, "skipped") ? BUS1_ERROR_STARTUP_SKIPPED : BUS1_ERROR_STARTUP_FAILURE);
                 if (r)
                         return error_trace(r);
         }
@@ -407,7 +402,7 @@ static int service_watch_unit_handler(sd_bus_message *message, void *userdata, s
          * though, as the job will never end up claiming its name.
          */
         if ((value && !strcmp(value, "failed")) || !condition_result) {
-                r = service_reset_activation(service);
+                r = service_reset_activation(service, BUS1_ERROR_UNIT_FAILURE);
                 if (r)
                         return error_trace(r);
         }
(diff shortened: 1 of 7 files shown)


src/launch/service.c
```diff
@@ -260,7 +254,13 @@ static int service_watch_jobs_handler(sd_bus_message *message, void *userdata, s
...
                  * can fail pending activations.
                  */
-                r = service_reset_activation(service);
+                _c_cleanup_(c_freep) char *error_message = NULL;
+
+                r = asprintf(&error_message, "job result: %s", result);
+                if (r < 0)
+                        return error_origin(-errno);
+
+                r = service_reset_activation(service, error_message);
```
teg, line 263: I think we should define this as something like `org.bus1.Error.<something matching the error we received>`.  If we also want to include a received human-readable error message (in case that's what systemd gives us), I think that's fine, but it should be secondary, I think.

src/launch/service.c
```diff
@@ -390,7 +390,7 @@ static int service_watch_unit_handler(sd_bus_message *message, void *userdata, s
          * anything to explicitly fail if they want to reset the activation.
          */
         if (value && !strcmp(value, "failed")) {
-                r = service_reset_activation(service);
+                r = service_reset_activation(service, "service failed");
```
teg, line 393: Ditto.

src/launch/service.c
```diff
@@ -532,7 +532,7 @@ static int service_start_unit_handler(sd_bus_message *message, void *userdata, s
                 }
         }
 
-        r = service_reset_activation(service);
+        r = service_reset_activation(service, "activation failed");
```
teg, line 535: Ditto.

teg (commented):
I like the overall idea, only request I have is to use well-defined error names, rather than human-readable strings.

src/launch/service.c
bluca: given these are strings, then we'll need to hard-code the currently known one and convert to the enum - is that ok?

dvdhrm: I would be fine with `org.bus1.DBus.Name.Error.StartupFailure` regardless of the exact type. But yeah, if we want it fine-grained, we need to add a strcmp() for each known result and convert it into a proper error-name.

dvdhrm (commented):
Some comments and suggestions inline. I like this, thanks for doing it!

I still wonder how to go about returning that information. I really dislike passing free-form sentences in our internal API (as @teg already mentioned), so there we should really refrain from it. However, how do we confer that information back to the caller? We have to use the pre-defined `org.freedesktop.DBus.Error.*` code, and we are only left with its own free-form field for additional information. Should be put our reverse-domain code there unmodified? This would allow callers to react to it. But it would misuse the field. Alternatively, we convert it into a free-form string as last step. Unsure what to do exactly. Comments?

src/launch/service.c
teg: Maybe I misunderstood, so take this with a grain of salt, but I would be against matching on free form text, but in favor of matching on reverse domain name error names if we have them to make it more finegrained.

teg:
> However, how do we confer that information back to the caller? We have to use the pre-defined `org.freedesktop.DBus.Error.*` code, and we are only left with its own free-form field for additional information. Should be put our reverse-domain code there unmodified? This would allow callers to react to it. But it would misuse the field. Alternatively, we convert it into a free-form string as last step. Unsure what to do exactly. Comments?

Hmm, I think I would vote for the second option. Not as useful, but the alternative is quite hacky.

bluca:
Ok, updated - I defined the errors as suggested, added enums and converters. PTAL!

teg (approved):
This looks great to me!

bluca:
@dvdhrm ping

bluca:
another ping @teg @dvdhrm

dvdhrm:
I am on 8month parental leave until March. I will defer any non-bugfixes until then. If you consider this PR critical, let me know and I will try to find some time for it.

bluca:
It's not super urgent, and if you are on parental leave you definitely have more important things going on :-)

bluca:
> I am on 8month parental leave until March. I will defer any non-bugfixes until then. If you consider this PR critical, let me know and I will try to find some time for it.

@dvdhrm ping, in case you are back to work

bluca:
@dvdhrm one more monthly ping. Anything else I should change here? Thanks!

bluca:
Rebased to fix merge conflict

dvdhrm:
I merged this manually. I did a follow-up commit which changes `bus1_error_*` to `controller_name_error_*`, to make sure we keep the namespacing. Additionally, I changed the signed type to `unsigned`, making value `0` be invalid. This is mostly to stay in line with our coding-style.

Thanks a lot for the patches! Very much appreciated!

bluca:
Thanks - FYI if you had left that as feedback, I'd have been happy to update the PR myself

dvdhrm:
> Thanks - FYI if you had left that as feedback, I'd have been happy to update the PR myself

I know, I know. But you already patiently waited for so long, so I didn't want to nag too much on coding style.

PR #278, opened 2021-12-04
launcher: use new StartUnitWithFlags method if available 

### LINKED ISSUE #276: On activation, skipped jobs are treated as successful
https://github.com/bus1/dbus-broker/blob/727bee57cd3e3b5806eb8599abbdd49984b75732/src/launch/service.c#L251

Shouldn't a `skipped` job be treated as a failed one and reset the activation? Currently we return `skipped` when the unit cannot be started. I'm thinking whether we should also report `skipped` when a condition fails. This came from a bug report from a user https://github.com/systemd/systemd/issues/21520

Also I just found that when you have a unit that is not running, and has a condition which is not verified, the dbus call will be stuck, and trying to call the dbus method again will do nothing, even if the condition in the meanwhile became true, until dbus-broker is restarted.

EG:

```
[Unit]
ConditionPathExists=/tmp/marker
[Service]
BusName=com.some.name
ExecStart=/usr/bin/some_service
```

Switching systemd to return JOB_SKIPPED when a condition fails, and treating `skipped` as a failure in the launcher snippet above, fixes the issue. But I'm not 100% sure it is the right thing to do? Was there a particular reason for treating `skipped` as successful activation?

bluca:
Check if the new method from systemd is available via Introspect(),
and if so use it.
This allows us to receive 'skipped' as the result when the service we
are trying to activate does not run because of a failed Condition
check, which would return a false 'done' with the older
StartUnit method, resulting in us just waiting for a unit to
show up that will never arrive.

Fixes #276

Requires new systemd interface from https://github.com/systemd/systemd/pull/21609 so opening as draft for now

### DIFF
--- a/src/launch/introspect.c
+++ b/src/launch/introspect.c
@@ -1,0 +1,96 @@
+/*
+ * Implements D-Bus introspection, allowing to check whether
+ * an object supports a given D-Bus interface. Used for backward
+ * compatibility checks.
+ */
+
+#include <expat.h>
+#include "launch/introspect.h"
+#include "util/error.h"
+#include "util/log.h"
+
+struct parse_params {
+    const char *interface_name;
+    bool *supported;
+};
+
+static void parser_begin_fn(void *userdata, const XML_Char *name, const XML_Char **attrs) {
+        struct parse_params *params = userdata;
+        const char *k, *v;
+
+        c_assert(params);
+        c_assert(params->interface_name);
+        c_assert(params->supported);
+
+        if (!attrs)
+                return;
+
+        if (strcmp(name, "method") != 0)
+                return;
+
+        while (*attrs) {
+                k = *(attrs++);
+                v = *(attrs++);
+
+                if (!k || !v)
+                        continue;
+
+                if (strcmp(k, "name") == 0 && strcmp(v, params->interface_name) == 0) {
+                        *params->supported = true;
+                        return;
+                }
+        }
+}
+
+int object_supports_interface(Launcher *launcher, const char *bus_name, const char *object, const char *interface_name) {
+        _c_cleanup_(sd_bus_error_free) sd_bus_error error = SD_BUS_ERROR_NULL;
+        _c_cleanup_(sd_bus_message_unrefp) sd_bus_message *reply_xml = NULL;
+        bool supported = false;
+        XML_Parser parser;
+        const char *xml;
+        int r;
+        struct parse_params params = {
+                .interface_name = interface_name,
+                .supported = &supported,
+        };
+
+        c_assert(launcher);
+        c_assert(bus_name);
+        c_assert(object);
+        c_assert(interface_name);
+
+        r = sd_bus_call_method(launcher->bus_regular,
+                               bus_name,
+                               object,
+                               "org.freedesktop.DBus.Introspectable",
+                               "Introspect",
+                               &error, &reply_xml, NULL);
+        if (r < 0)
+                return error_origin(r);
+
+        r = sd_bus_message_read(reply_xml, "s", &xml);
+        if (r < 0)
+                return error_origin(r);
+
+        parser = XML_ParserCreate(NULL);
+        if (!parser)
+                return error_origin(-ENOMEM);
+
+        XML_SetUserData(parser, &params);
+        XML_SetElementHandler(parser, parser_begin_fn, NULL);
+
+        r = XML_Parse(parser, xml, strlen(xml), 1);
+        if (r != XML_STATUS_OK) {
+                log_append_here(&launcher->log, LOG_ERR, 0, NULL);
+                r = log_commitf(&launcher->log,
+                                "Parsing the Introspect XML of '%s' failed: %s\n",
+                                object,
+                                XML_ErrorString(XML_GetErrorCode(parser)));
+                if (r)
+                        return error_fold(r);
+
+                return error_origin(-EINVAL);
+        }
+
+        return supported;
+}
--- a/src/launch/launcher.c
+++ b/src/launch/launcher.c
@@ -18,6 +18,7 @@
 #include "catalog/catalog-ids.h"
 #include "dbus/protocol.h"
 #include "launch/config.h"
+#include "launch/introspect.h"
 #include "launch/launcher.h"
 #include "launch/nss-cache.h"
 #include "launch/policy.h"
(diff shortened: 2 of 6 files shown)


bluca:
https://github.com/systemd/systemd/pull/21609 has been approved and will be merged soon and be part of the v250 release, so this is now ready for review. PTAL!

src/launch/service.c
```diff
@@ -539,11 +540,104 @@ static int service_start_unit_handler(sd_bus_message *message, void *userdata, s
...
         int r;
 
+        /*
+         * When a Condition*= setting fails, the JobRemoved signal normally returns
+         * 'done' which we interpret as a success, but it means we will wait in vain
+         * for the service to start. It cannot be changed without breaking backward
+         * compatibility, so a new StartUnitWithFlags was added to change it on-demand.
+         * Let's use it if it is available, and fallback to the old StartUnit otherwise.
+         */
+        r = start_methods_with_flags_supported(launcher, &use_new_interface);
```
teg, line 637: Would it make sense to just try with the new interface, and then fallback to the old one if it fails (with the right error code)? We'd only try once of course, and remember the decision.

teg (commented):
This makes sense to me! I don't love the introspection, so made a suggestion inline, but maybe that's just me. Let's hear what @dvdhrm has to say.

src/launch/service.c
bluca: I thought about that, but given it's all async, you wouldn't get a response immediately, so handling it properly you'd need to queue back the service start request from the response handler as a special case, and a lot of extra effort in general. Given dbus supports introspection exactly for this reason, and there's already an xml library being used in the project, it seemed to me to be much easier to go down this route. It's only done once and cached.

If the xml parsing is the problem, I could "hack" it a bit given we just need to know if the method is present, so in theory a simple check of the response xml string to see if it contains the "StartUnitWithFlags" substring would be enough to know, without full XML parsing. Would that be better?

teg: Yeah, that's a fair point. And if we are going to introspect, then we should absolutely do the full XML parsing as you are doing (though maybe it would be cleaner to have the code in a separate file, the same way we deal with loading the policy?).

I don't see any deadlock issues in this implementation, though my gut feeling is that blocking calls while handling async calls is a bit scary, and if we add more of these it might get convoluted quickly.

Would it be more predictable to move this into `launcher_run()`, maybe introducing a new helper function `launcher_introspect()` which could be extended in the future to read out more values, but for now just sets a boolean in the `Launcher` object telling us which interface to use?

bluca: Yeah, I can look into moving it at startup, sounds like a good idea, will do so tomorrow

bluca: Ok, updated as requested. Ptal!

src/launch/introspect.c
```diff
@@ -0,0 +1,96 @@
...
+        parser = XML_ParserCreate(NULL);
+        if (!parser)
+                return error_origin(-ENOMEM);
+
+        XML_SetUserData(parser, &params);
+        XML_SetElementHandler(parser, parser_begin_fn, NULL);
+
+        r = XML_Parse(parser, xml, strlen(xml), 1);
+        if (r != XML_STATUS_OK) {
+                log_append_here(&launcher->log, LOG_WARNING, 0, NULL);
```
teg, line 84: I think we should make this a hard failure, so log this at ERROR level. If my memory serves me right I think we only ignore errors for the sake of backwards compatibility, but we don't expect systemd to pass us invalid XML.

src/launch/introspect.c
```diff
@@ -0,0 +1,96 @@
...
+                r = log_commitf(&launcher->log,
+                                "Parsing the Introspect XML of '%s' failed, ignoring: %s\n",
```
teg, line 86: If we change as above, we would just skip the "ignoring".

src/launch/introspect.c
```diff
@@ -0,0 +1,96 @@
...
+                                object,
+                                XML_ErrorString(XML_GetErrorCode(parser)));
+                if (r)
+                        return error_fold(r);
+
+                return -EINVAL;
```
teg, line 92: This should be `error_origin(-EINVAL)`, see `util/error.{h,c}` for details.

src/launch/launcher.c
```diff
@@ -1434,6 +1435,22 @@ int launcher_run(Launcher *launcher) {
...
+         * for the service to start. It cannot be changed without breaking backward
+         * compatibility, so a new StartUnitWithFlags was added to change it on-demand.
+         * Let's use it if it is available, and fallback to the old StartUnit otherwise.
+         */
+        r = object_supports_interface(launcher,
+                                      "org.freedesktop.systemd1",
+                                      "/org/freedesktop/systemd1",
+                                      "StartUnitWithFlags");
+        if (r < 0)
+                error_origin(r); /* Log but continue with backward-compatible method */
```
teg, line 1450: This should be `return error_fold(r)`. If we wanted to make it a soft error, we should define and return a positive error code (see `util/error.c`), but as mentioned above, I think we should make this a hard error, so that point is moot.

teg (changes requested):
This is looking really good! I had some minor points about error handling inline (our error handling is a bit non-standard).
Otherwise, I'm pretty happy with this, though would love to wait for systemd to hit rawhide, so we can test that this works as expected with the new interface.

src/launch/launcher.c
bluca: This means we'll also fail to run in case there's a problem with connecting to the bus and calling Introspectable, not just with malformed XML. Is that safe to do? I don't mind either way, was just trying to play it on the safe side, given it's backward compatible.

bluca: updated as requested for now, if you change your mind let me know and I'll change it back

bluca:
> This is looking really good! I had some minor points about error handling inline (our error handling is a bit non-standard). Otherwise, I'm pretty happy with this, though would love to wait for systemd to hit rawhide, so we can test that this works as expected with the new interface.

I've tested this by building an image with mkosi, it's really simple if you want to give it a shot. Simply install mkosi (it's available as a package in fedora), clone the systemd git tree on the latest main branch, and from the systemd git repo directory run:

```
$ mkdir -p mkosi.skeleton/usr/bin/
$ cp <path/to/dbus/broker/build/dir>/dbus-broker <path/to/dbus/broker/build/dir>/dbus-broker-launch mkosi.skeleton/usr/bin/
$ sudo mkosi build
$ sudo mkosi boot
```

This will build an image of the current running distro (so Fedora in your case) and add the systemd binaries built from the repo, and then copy the dbus broker binaries on top of it, and then boot it using nspawn. To check the failure mode, simply add a drop-in for any bus-activatable unit with:

```
[Unit]
ConditionPathExists=/tmp/marker
```

You can verify that before creating the marker file the unit doesn't activate and you get an error immediately with this patch, but it gets stuck without.

src/launch/launcher.c
teg: I figure if we cannot talk to pid1 we'll have bigger problems (and we try to err on the side of failing early as a general rule). Though let's see if @dvdhrm has any other opinions.

teg (approved):

src/launch/launcher.c
bluca: this also affects user sessions and buses, not just pid1/system bus (if it makes any difference)

teg:
This is looking great to me! Though I'll give @dvdhrm a chance to have a look before merging.

bluca:
The interface is going to be reworked a bit, so closing for now

PR #282, opened 2022-01-26
ci: introduce CodeQL

mrc0mmand:
This commit introduces GitHub's CodeQL Action[0] which regularly scans
code for possible errors and vulnerabilities. I borrowed our custom
configs from the systemd repository, since they enable a couple of other
security-related checks which are disabled by default.

[0] https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning

---

The results are then shown in the PR (if the PR introduces new alerts) or in the Security tab for the general overview, e.g.:

https://github.com/mrc0mmand/dbus-broker/security/code-scanning?query=+pr%3A1+
or
https://github.com/systemd/systemd/security/code-scanning

One minor issue is that the CodeQL workflow processes files from submodules, but it doesn't show preview for them in the UI, e.g.:

image

However, the offending line is still discoverable from the alerts overview:

image

so it's not completely useless.

Anyway, this is just a proposal, if this is something which you don't deem useful, feel free to close this PR.

### DIFF
--- a/.github/codeql-config.yml
+++ b/.github/codeql-config.yml
@@ -1,0 +1,9 @@
+---
+# vi: ts=2 sw=2 et:
+name: "CodeQL config"
+
+disable-default-queries: false
+
+queries:
+  - name: Enable possibly useful queries which are disabled by default
+    uses: ./.github/codeql-custom.qls
--- a/.github/codeql-custom.qls
+++ b/.github/codeql-custom.qls
@@ -1,0 +1,35 @@
+---
+# vi: ts=2 sw=2 et syntax=yaml:
+#
+# Note: it is not recommended to directly reference the respective queries from
+#       the github/codeql repository, so we have to "dance" around it using
+#       a custom QL suite
+# See:
+#   - https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#running-additional-queries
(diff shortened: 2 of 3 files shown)


dvdhrm:
I fixed the 2 signed-overflow bugs in c-dvar. Thanks for catching them!

Regarding adding CodeQL: I added the `scan` branch for now, with this pulled in. However, our new build-instructions use meson-0.60, which is not available on the standard github runners. I will defer merging this until this is included, or until I figured out a way to build on a fedora container, but then run codeql on the result but on the github-runner.

mrc0mmand:
I updated the setup code to install both meson and ninja from pip. It's not ideal, but it seems to be working, at least according to https://github.com/mrc0mmand/dbus-broker/pull/1.

dvdhrm:
Right, we can just `pip install ...` it. I merged this now with some minor reworks. Thanks a lot! Any further changes we can now discuss in other PRs. I just wanted to make sure the base setup is merged and running. So again, thanks a lot!

evverx:
Note to self: the autobuild part should be replaced with `meson -Daudit=true -Dselinux=true -Dapparmor=true build && ninja -C build` to get CodeQL to analyze the optional dependencies instead of the fallbacks. Unlike systemd dbus-broker doesn't auto-detect the dependencies so autobuild can't catch them itself.

PR #285, opened 2022-03-12
Fix hang when trying to activate a systemd service with unsatisfied conditions

### LINKED ISSUE #276: On activation, skipped jobs are treated as successful

msizanoen1:
When systemd is requested to start a service with unsatisfied conditions, it will report that the service was started successfully but in reality the service was never started. This causes the client trying to call a method on a DBus-activated service to hang or timeout instead of receiving an error.

This PR fix this by checking for the `ConditionResult` property on the service unit object and fail activation if it's `false`.

How to reproduce:
 - Get a Fedora Workstation 35 machine with no Bluetooth (or a VM)
 - Type: `busctl call org.bluez / org.freedesktop.DBus.Peer Ping`

Expected result: `Call failed: Could not activate remote peer.`

Actual result: The command times out.

Fixes #276

### DIFF
--- a/src/launch/service.c
+++ b/src/launch/service.c
@@ -398,20 +398,19 @@ static int service_watch_unit_handler(sd_bus_message *message, void *userdata, s
         return 0;
 }
 
-static int service_watch_unit(Service *service, const char *unit) {
-        _c_cleanup_(c_freep) char *object_path = NULL;
+static int service_watch_unit_callback(sd_bus_message *message, void *userdata, sd_bus_error *errorp) {
+        Service *service = userdata;
+        const char *object_path;
         int r;
 
-        assert(!service->slot_watch_unit);
+        if (sd_bus_message_get_error(message))
+                goto error;
 
-        r = sd_bus_path_encode(
-                "/org/freedesktop/systemd1/unit",
-                unit,
-                &object_path
-        );
-        if (r < 0)
-                return error_origin(r);
+        r = sd_bus_message_read(message, "o", &object_path);
+        if (error_trace(r) < 0)
+                goto error;
 
+        service->slot_watch_unit = sd_bus_slot_unref(service->slot_watch_unit);
         r = sd_bus_match_signal_async(
                 service->launcher->bus_regular,
                 &service->slot_watch_unit,
@@ -422,6 +421,33 @@ static int service_watch_unit(Service *service, const char *unit) {
                 service_watch_unit_handler,
                 NULL,
                 service
+        );
+        if (error_trace(r) < 0)
+                goto error;
+
+        return 1;
+
+error:
+        service_reset_activation(service);
+        return 1;
+}
+
+static int service_watch_unit(Service *service, const char *unit) {
+        int r;
+
+        assert(!service->slot_watch_unit);
+
+        r = sd_bus_call_method_async(
+                service->launcher->bus_regular,
+                &service->slot_watch_unit,
+                "org.freedesktop.systemd1",
+                "/org/freedesktop/systemd1",
+                "org.freedesktop.systemd1.Manager",
+                "LoadUnit",
+                service_watch_unit_callback,
+                service,
+                "s",
+                unit
         );
         if (r < 0)
                 return error_origin(r);
--- a/src/launch/service.c
+++ b/src/launch/service.c
@@ -294,6 +294,7 @@ static int service_watch_jobs(Service *service) {
 static int service_watch_unit_handler(sd_bus_message *message, void *userdata, sd_bus_error *errorp) {
         Service *service = userdata;
         const char *interface = NULL, *property = NULL, *value = NULL;
+        int condition_result = 1;
         int r;
 
         /*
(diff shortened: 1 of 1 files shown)


src/launch/service.c
```diff
@@ -399,19 +399,30 @@ static int service_watch_unit_handler(sd_bus_message *message, void *userdata, s
...
+        _c_cleanup_(sd_bus_message_unrefp) sd_bus_message *m = NULL;
         int r;
 
         assert(!service->slot_watch_unit);
 
-        r = sd_bus_path_encode(
-                "/org/freedesktop/systemd1/unit",
-                unit,
-                &object_path
+        r = sd_bus_call_method(
```
bluca, line 423: other calls into pid1 here are async - are we sure it's ok to block on this one?

msizanoen1: `org.freedesktop.systemd1.Manager.LoadUnit` should at most cause a single filesystem access in pid1, so it should complete within a reasonably short amount of time.

dvdhrm: You cannot do blocking calls into systemd here. This is not about the costs, but rather about circular dependencies. You easily run into deadlocks, like if systemd tries to call into polkit before answering the request. There is so many possibilities how this can go wrong and dbus-broker ends up waiting on systemd, which ends up waiting on D-Bus to serve the request.

Preferably, I would not to that call to LoadUnit(), unless really necessary. It definitely is not trivial and I would prefer a separate PR and discussion, because the other fix looks fine and could be merged.

msizanoen1: Would just converting the call to asynchronous be sufficient?

msizanoen1: Seems like it would be possible to get the unit object path from the `org.freedesktop.systemd1.Job` object instead.

bluca (approved):

src/launch/service.c
```diff
@@ -398,20 +413,19 @@ static int service_watch_unit_handler(sd_bus_message *message, void *userdata, s
         return 0;
 }
 
-static int service_watch_unit(Service *service, const char *unit) {
-        _c_cleanup_(c_freep) char *object_path = NULL;
+static int service_watch_unit_callback(sd_bus_message *message, void *userdata, sd_bus_error *errorp) {
```
dvdhrm, line 416: Maybe `service_watch_unit_load_handler()` to be consistent with our naming-scheme?

src/launch/service.c
```diff
@@ -398,20 +413,19 @@ static int service_watch_unit_handler(sd_bus_message *message, void *userdata, s
...
-        r = sd_bus_path_encode(
-                "/org/freedesktop/systemd1/unit",
-                unit,
-                &object_path
-        );
-        if (r < 0)
-                return error_origin(r);
+        r = sd_bus_message_read(message, "o", &object_path);
+        if (error_trace(r) < 0)
+                goto error;
```
dvdhrm, line 426: This should return with `error_origin(r)` as the fatal error originates here.

src/launch/service.c
```diff
@@ -398,20 +413,19 @@ static int service_watch_unit_handler(sd_bus_message *message, void *userdata, s
...
 
+        service->slot_watch_unit = sd_bus_slot_unref(service->slot_watch_unit);
```
dvdhrm, line 428: This should probably be at the top of the function to clean everything up and prepare for any followup. I don't think anything leaks if doing this late, but I think it would be easier to review if this is done early.

src/launch/service.c
```diff
@@ -423,6 +437,33 @@ static int service_watch_unit(Service *service, const char *unit) {
                 NULL,
                 service
         );
+        if (error_trace(r) < 0)
+                goto error;
```
dvdhrm, line 441: Again, as above, this should return `error_origin(r)`.

src/launch/service.c
```diff
@@ -423,6 +437,33 @@ static int service_watch_unit(Service *service, const char *unit) {
...
+                &service->slot_watch_unit,
+                "org.freedesktop.systemd1",
+                "/org/freedesktop/systemd1",
+                "org.freedesktop.systemd1.Manager",
+                "LoadUnit",
+                service_watch_unit_callback,
+                service,
+                "s",
+                unit
+        );
```
dvdhrm, line 466: What happens if systemd reloads configuration between our call to `LoadUnit` and the following watch-operation? That is, if the unit in question changes exactly at that moment, we will possible watch the wrong path, right? Not a big issue, as this is inherent to the way systemd-reloads work, but I was just wondering what the story there is.

Oh, and same issue arises if the unit is modified at any time while we watch it. We will never re-resolve the object-path, so any changes to aliases will be ignored by us. Not really related to your patch, as the old code already suffered from those problems, but definitely something to keep in mind.

dvdhrm (approved):
This looks good! Some minor corrections inline, which I took the liberty to fix myself.

I applied the patches manually! Thanks a lot!

Issue #169, opened 2018-10-10
AppArmor support


bitstreamout:
I see this here in the journal
```
 dbus-broker-launch[1035]: AppArmor enabled, but not supported. Ignoring.
```

dvdhrm:
The AppArmor code is, to a large extent, a downstream patchset carried by Ubuntu in their respective packages. The upstream linux kernel does not have the required AppArmor features to implement AppArmor-support in dbus-broker. In particular, `SO_PEERSEC` is not sufficiently implemented.

`dbus-daemon(1)` carries patches against the downstream AppArmor code of Ubuntu. We currently are a bit reluctant to add code that relies on downstream kernel features. However, we did adjust the dbus-broker infrastructure to easily allow downstream patches that add AppArmor support. Meaning, we already parse all the AppArmor related configuration and can easily insert the required AppArmor hooks.

bitstreamout:
Hmm ... the apparmor patch set is part of our kernel here.  Otherwise I would not see this messages I guess.  I found https://lkml.org/lkml/2018/5/4/476 ... now let's see if your patchset will reach the kernel

dvdhrm:
Basic Apparmor support is upstream in the linux kernel. If you now enable apparmor support in your dbus configuration, you will see the message you mentioned. The are two modes to run dbus-AppArmor in: enforcing or informative. Your configuration apparently uses the latter, hence AppArmor is ignored by dbus-broker, as it is informative only. If you selected the enforcing mode, dbus-broker would refuse to start (since it lacks apparmor support).

If anyone wants *enforcing* behavior of AppArmor in dbus-broker, they have to patch it into dbus-broker manually, or upstream the AppArmor patches so we can work on it.

The patchset you linked to is not related to this issue. It was already merged upstream and is about `SO_PEERSEC` support on socketpairs.

dvdhrm:
For documentational purposes:

The downstream Ubuntu patch that currently provides the required kernel functionality is:

    https://git.launchpad.net/~ubuntu-kernel/ubuntu/+source/linux/+git/cosmic/commit/security/apparmor?id=2d33c63704a49b005dfa4597882be265ce0bdb06

I saved it on github as well:

    https://gist.github.com/dvdhrm/0b8c556960957970d646c9bba814692a

In particular, we need upstream kernel sources to include:

```
+	LSM_HOOK_INIT(unix_stream_connect, apparmor_unix_stream_connect),
[...]

PR #286, opened 2022-03-18
Apparmor support

sre:
Hi,

This adds support for AppArmor using libapparmor and the Ubuntu kernel patch
requested in #169. For the implementation dbus-daemon has been used as reference.
The patchset handles restriction of all common DBus operations (sending, receiving,
bus owning and monitoring) - just like dbus-daemon.

Note, that the downstream kernel patch is no longer limited to Ubuntu. It has been
applied to some kernels used in the embedded sector. Missing support for AppArmor
is the limiting factor to switch from dbus-daemon to dbus-broker on these embedded
systems and the reason I wrote this code.

Regarding the status of the kernel patch required for this support: The main reason,
that the kernel is still missing support for kernel based af_unix/dbus meditation is a
pending code restructuring that got postponed for multiple years. Current expectation
is, that this restructuring finally happens in the 5.19 cycle (*fingers crossed*).

-- Sebastian


src/bus/policy.h
```diff
@@ -162,16 +166,21 @@ int policy_snapshot_dup(PolicySnapshot *snapshot, PolicySnapshot **newp);
 int policy_snapshot_check_connect(PolicySnapshot *snapshot);
 int policy_snapshot_check_own(PolicySnapshot *snapshot, const char *name);
 int policy_snapshot_check_send(PolicySnapshot *snapshot,
-                               const char *subject_context,
+                               const char *sender_context,
```
dvdhrm, line 169: The sender context should be taken from `snapshot` (see `snapshot->seclabel`). Otherwise, you will query the context too late. Note that the point of the policy-snapshot is to snapshot all sender information at a particular time, and then *later* check whether it matches. This is used for activation messages where sender and receiver information is *not* available simultaneously. If this does not work for apparmor, this needs to be somehow documented.

src/util/apparmor.c
```diff
@@ -93,3 +163,489 @@ int bus_apparmor_dbus_supported(bool *supportedp) {
...
+
+        va_start(ap, fmt);
+        r = vasprintf(&message, fmt, ap);
+        va_end(ap);
+        if (r < 0)
+                return r;
+
+        /* XXX: we don't have access to any context, so can't find
+         * the right UID to use, follow dbus-daemon(1) and use our
+         * own. */
```
dvdhrm, line 195: Why can't we pass the context in? I don't understand this comment. Which context would be needed?

dvdhrm (changes requested):
Thanks a lot for doing this! Some basic questions on the code inline, but it generally it looks good. I intend to merge this when I find the time to properly review this. But it will not make the next release, as I really need to tag that as soon as possible.

The second patch is quite big and could easily be split in lots of smaller ones. This would make reviewing a lot easier and make me more confident I don't miss anything. I can pick the patch apart myself and apply the changes individually, if you prefer. But if you want to do that yourself,  please go ahead. I don't mind either way.

Again, thanks for looking into this!

src/util/apparmor.c
sre: I took over that function incl. the comment from the dbus-broker SELinux implementation without investigating further (and adapted it for apparmor).

sre:
Thanks for having a look at it. I will check the things you pointed out next week and also try to figure out a way to split the second patch into smaller chunks.

Conan-Kudo:
@sre Is there a link to the patch review on lore for the revised patch slated for 5.19 yet?

sre:
> @sre Is there a link to the patch review on lore for the revised patch slated for 5.19 yet?

No, not that I'm aware of.

meson.build
```diff
@@ -48,6 +48,15 @@ if use_audit
...
 
+#
+# Config: apparmor
+#
+
+use_apparmor = get_option('apparmor')
+if use_apparmor
+        dep_libapparmor = dependency('libapparmor', version: '>=3.0')
+endif
+
```
dvdhrm, line 59: I moved this above the `audit`-check, since `apparmor` is before `audit` in alphabetical order ;)

meson_options.txt
```diff
@@ -1,3 +1,4 @@
+option('apparmor', type: 'boolean', value: false, description: 'AppArmor support')
 option('audit', type: 'boolean', value: false, description: 'Audit support')
```
dvdhrm, line 2: Here it was in the correct order :)

src/bus/driver.c
```diff
@@ -2337,6 +2347,17 @@ static int driver_dispatch_method(Peer *peer, const DriverMethod *methods, uint3
...
 }
 
+static int driver_map_denied_error(int err) {
+        switch (err) {
+        case POLICY_E_APPARMOR_ACCESS_DENIED:
+                return BUS_LOG_POLICY_TYPE_APPARMOR;
+        case POLICY_E_SELINUX_ACCESS_DENIED:
+                return BUS_LOG_POLICY_TYPE_SELINUX;
+        default:
+                return BUS_LOG_POLICY_TYPE_INTERNAL;
```
dvdhrm, line 2357: I did some re-ordering all over the place to stay consistent with the order the enum is defined. So first check INTERNAL (if it applies), then SELINUX, and APPARMOR last. Not a big deal, just for consistency.

src/bus/driver.c
```diff
@@ -2352,24 +2373,29 @@ static int driver_dispatch_interface(Peer *peer, uint32_t serial, const char *in
...
 
-        r = policy_snapshot_check_send(peer->policy, NULL, NULL, interface, member, path, message->header->type, false, message->metadata.fields.unix_fds);
+        r = policy_snapshot_check_send(peer->policy, NULL, NULL, 0, interface, member, path, message->header->type, false, message->metadata.fields.unix_fds);
         if (r) {
-                if (r == POLICY_E_ACCESS_DENIED || r == POLICY_E_SELINUX_ACCESS_DENIED) {
+                switch (r) {
+                case POLICY_E_ACCESS_DENIED:
+                case POLICY_E_APPARMOR_ACCESS_DENIED:
+                case POLICY_E_SELINUX_ACCESS_DENIED:
+			{
```
dvdhrm, line 2382: I know we have a mixed style of switch and if clauses, but I think I prefer the if-clause here. I fixed it up everywhere. I also added the same errno-check to all instances of `policy_snapshot_check_{send,receive}`, you missed one in driver.c. I also changed it to always check for SELINUX as well. Technically, SELINUX only checks for SEND, not RECEIVE, but the callsite should be consistent, imho. No reason to be overly pedantic there, it just makes it a lot harder to argue about it, and nothing really technically prevents us from adding selinux checks to recv as well.

src/bus/policy.c
```diff
@@ -688,11 +695,13 @@ int policy_registry_import(PolicyRegistry *registry, CDVar *v) {
                         return error_fold(r);
         }
 
-        c_dvar_read(v, "]b)>", &apparmor);
+        c_dvar_read(v, "]bs)>", &apparmor, &bustype);
```
dvdhrm, line 698: I also fixed up `test/dbus/util-broker.c` so the test-suite should now correctly set the bustype as well.

src/launch/policy.h
```diff
@@ -114,6 +115,7 @@ struct Policy {
                 .gid_tree = C_RBTREE_INIT,                                              \
                 .selinux_list = C_LIST_INIT((_x).selinux_list),                         \
                 .apparmor_mode = CONFIG_APPARMOR_ENABLED,                               \
+                .bus_type = CONFIG_BUS_TYPE_SYSTEM,                                     \
```
dvdhrm, line 118: Is `system` actually the default? The man-page doesn't mention it, and I didn't check the dbus sources. Did you?

src/util/apparmor.c
```diff
@@ -57,3 +69,489 @@ int bus_apparmor_is_enabled(bool *enabledp) {
...
+int bus_apparmor_dbus_supported(bool *supportedp) {
+        _c_cleanup_(c_fclosep) FILE *f = NULL;
+        char buffer[LINE_MAX] = {};
+        bool supported;
+
+        f = fopen("/sys/kernel/security/apparmor/features/dbus/mask", "re");
+        if (f) {
+                errno = 0;
+                if (!fgets(buffer, sizeof(buffer), f)) {
+                        if (ferror(f) && errno != EINVAL)
```
dvdhrm, line 91: I assume this EINVAL check was copy-pasted. Looks like an error. I dropped it. If it was intentional, let me know! I also dropped it from the old `is_enabled()` code.

dvdhrm (commented):
So I pulled in most of the preparatory patches, and I adjusted some bits. The major apparmor.c code is still missing. I am working on it.

I commented on all things I changed. Let me know if something is unclear, or if I unintentionally broke something.

src/launch/policy.h
sre: No, I did not check when writing this and totally forgot about it. I just did check dbus-daemon. Apparently it supports arbitrary bus types incl. the bus type being empty and then has special handling for "session" and "system". So parsing needs to go and instead the string should be used directly.

sre:
> So I pulled in most of the preparatory patches, and I adjusted some bits. The major apparmor.c code is still missing. I am working on it.

Thanks, the changes look fine to me.

> Let me know if something is unclear, or if I unintentionally broke something.

I just rebased the pull-request to current HEAD dropping the merged patches (no other changes) and the branch is still working. I will look into the CI issues and the bus_type problem.

src/launch/policy.h
dvdhrm: Yeah, I think we need to avoid parsing it and just send it through, indeed.

src/bus/policy.h
```diff
@@ -173,6 +173,8 @@ int policy_snapshot_check_send(PolicySnapshot *snapshot,
                                bool broadcast,
                                size_t n_fds);
 int policy_snapshot_check_receive(PolicySnapshot *snapshot,
+                                  const char *sender_seclabel,
+                                  const char *receiver_seclabel,
```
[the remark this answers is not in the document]
dvdhrm: I dropped it myself and pushed the patch!

src/bus/policy.c
```diff
@@ -691,9 +697,11 @@ int policy_registry_import(PolicyRegistry *registry, CDVar *v) {
...
         c_dvar_read(v, "]bs)>", &apparmor, &bustype);
 
-        if (apparmor)
-                /* XXX: AppArmor is currently not supported. */
-                return POLICY_E_INVALID;
+        if (apparmor) {
+                r = bus_apparmor_set_bus_type(registry->apparmor, bustype);
+                if (r)
+                        return error_trace(r);
+        }
```
dvdhrm, line 704: We technically want the launcher and broker to be independent. Imagine they are in different projects, and the launcher has apparmor enabled, but the broker has not. In this case, the launcher would forward the information to the broker even in `enforced/required` mode, but the broker would then silently ignore it because the `bus_apparmor_set_bus_type()` fallback just returns success unconditionally.

Maybe this is a misconfiguration, and the launcher needs to ensure that the broker it runs with has apparmor enabled. But in this case, there is no point in passing the `apparmor-boolean` at all. We can just always call `bus_apparmor_set_bus_type()` and drop the boolean entirely. I think I would go with that initially, and rely on launchers to depend on correctly compiled/configured brokers.

dvdhrm: Note that SELinux does not suffer from the same problem, because it does not have the `required` mode as part of dbus-configuration. It simply relies on a correctly configured system. Though, it does have the `if_selinux_enabled` condition in configuration parsing, which again is really weird if the launcher and broker compile-time-configuration mismatch. Umpf.

dvdhrm: Ahh, no, I am wrong. If apparmor=disabled, we don't want to call set_bus_type(), I guess? This is weird. It means you need to check for `bustype == NULL` in your `*_check_*()` functions, and return success if no bustype is set (because it means apparmor was not configured. Right?

src/bus/policy.c
```diff
@@ -777,6 +786,7 @@ PolicySnapshot *policy_snapshot_free(PolicySnapshot *snapshot) {
         while (snapshot->n_batches-- > 0)
                 policy_batch_unref(snapshot->batches[snapshot->n_batches]);
         free(snapshot->seclabel);
+        bus_apparmor_registry_unref(snapshot->apparmor);
         bus_selinux_registry_unref(snapshot->selinux);
```
dvdhrm, line 790: Again, wrong ordering.

src/bus/policy.c
```diff
@@ -1082,6 +1114,16 @@ int policy_snapshot_check_receive(PolicySnapshot *snapshot,
                                   size_t n_fds) {
         PolicyVerdict verdict = POLICY_VERDICT_INIT;
         size_t i;
+        int r;
+
+        r = bus_apparmor_check_xmit(snapshot->apparmor, false, sender_seclabel, receiver_seclabel,
```
dvdhrm, line 1119: This is now: `subject_seclabel, snapshot->seclabel`, rather than `sender_seclabel, receiver_seclabel`

src/launch/launcher.c
```diff
@@ -1197,18 +1197,25 @@ static int launcher_reload_config(Launcher *launcher) {
...
+                } else if (enabled) {
+                        policy.apparmor_mode = CONFIG_APPARMOR_ENABLED;
+                } else {
+                        policy.apparmor_mode = CONFIG_APPARMOR_DISABLED;
+                }
 
-                policy.apparmor_mode = CONFIG_APPARMOR_DISABLED;
                 break;
         }
         case CONFIG_APPARMOR_REQUIRED:
```
dvdhrm, line 1221: You missed the REQUIRED part?

src/launch/launcher.c
```diff
@@ -1359,22 +1366,35 @@ int launcher_run(Launcher *launcher) {
...
+                r = bus_apparmor_dbus_supported(&supported);
+                if (r)
+                        return error_fold(r);
+
+                if (!enabled || !supported) {
+                        fprintf(stderr, "AppArmor required, but not supported. Exiting.\n");
+                        return 0;
+                }
+
+                policy.apparmor_mode = CONFIG_APPARMOR_ENABLED;
```
dvdhrm, line 1396: maybe turn this entire thing into a helper function? We duplicate it now in initial-load and reload.

src/util/apparmor.c
```diff
@@ -10,11 +10,23 @@
...
 #include <c-stdaux.h>
+#include <libaudit.h>
 #include <stdlib.h>
 #include <stdio.h>
+#include <sys/apparmor.h>
 #include "util/apparmor.h"
+#include "util/audit.h"
 #include "util/error.h"
+#include "util/ref.h"
+#include "bus/name.h"
```
dvdhrm, line 23: "bus/name.h" should be before "util/*.h"

src/util/apparmor.c
```diff
@@ -10,11 +10,23 @@
...
+
+struct BusAppArmorRegistry {
+        _Atomic unsigned long n_refs;
+        const char *fallback_context;
+        const char *bustype;
```
dvdhrm, line 28: I think both should just drop `const`. Definitely for `bustype`, and I think even for `fallback_context` it would be better to drop it.

src/util/apparmor.c
```diff
@@ -93,3 +105,453 @@ int bus_apparmor_dbus_supported(bool *supportedp) {
...
+        int ret;
+        bool enabled;
+
+        ret = bus_apparmor_dbus_supported(&enabled);
+        if (ret || !enabled)
+                return false;
+
+        ret = bus_apparmor_is_enabled(&enabled);
+        if (ret || !enabled)
+                return false;
```
dvdhrm, line 119: You are calling this on every message transaction. I don't think we can do that, that is quite slow. I think it is better to just check `registry->bustype` and consider it disabled if it is NULL.

src/util/apparmor.c
```diff
@@ -93,3 +105,453 @@ int bus_apparmor_dbus_supported(bool *supportedp) {
...
+        memcpy((char *)registry->fallback_context, fallback_context, n_fallback_context);
+
+        *registryp = registry;
+        registry = NULL;
+        return 0;
+}
+
+static void bus_apparmor_registry_free(_Atomic unsigned long *n_refs, void *userdata) {
+        BusAppArmorRegistry *registry = c_container_of(n_refs, BusAppArmorRegistry, n_refs);
+
```
dvdhrm, line 171: `free(registry->bustype)`?

src/util/apparmor.c
```diff
@@ -93,3 +105,453 @@ int bus_apparmor_dbus_supported(bool *supportedp) {
...
+}
+
+BusAppArmorRegistry *bus_apparmor_registry_unref(BusAppArmorRegistry *registry) {
+        if (registry)
+                ref_dec(&registry->n_refs, bus_apparmor_registry_free, NULL);
+
+        return NULL;
+}
+
+int bus_apparmor_set_bus_type(BusAppArmorRegistry *registry, const char* bustype) {
```
dvdhrm, line 189: Maybe throw in a `c_assert(!registry->bustype)`

src/util/apparmor.c
```diff
@@ -93,3 +105,453 @@ int bus_apparmor_dbus_supported(bool *supportedp) {
...
+        registry->bustype = strdup(bustype);
+        if (!registry->bustype)
+                return -ENOMEM;
```
dvdhrm, line 192: `return error_origin(-ENOMEM)`

src/util/apparmor.c
```diff
@@ -93,3 +105,453 @@ int bus_apparmor_dbus_supported(bool *supportedp) {
...
+                                return r;
+                        if (audit_tmp)
+                                *audit = 1;
+                }
+        } else if (nameset->type != NAME_SET_TYPE_EMPTY) {
+                c_assert(0);
+                r = -EINVAL;
+        }
+
+        return r;
```
dvdhrm, line 377: Uff. This is slow. I mean, we optimized everything so we have a single allocation per message transaction. Now the apparmor code adds one allocation per name the receiver and sender own. I am fine keeping the code, but man those LSMs really drag things down. Makes me a bit sad :-/

src/util/apparmor.c
```diff
@@ -93,3 +105,453 @@ int bus_apparmor_dbus_supported(bool *supportedp) {
...
+        _c_cleanup_(c_freep) char *qstr = NULL;
+        int r;
+        /* the AppArmor API uses pointers to int for pointers to boolean */
+        int allow = false, audit = true;
+
+        if (!is_apparmor_enabled())
+                return 0;
+
+        if (!condup)
+                return -ENOMEM;
```
dvdhrm, line 410: `error_origin(-ENOMEM)`, and I would suggest moving the `strdup()` out of the variable declaration.

src/util/apparmor.c
```diff
@@ -93,3 +105,453 @@ int bus_apparmor_dbus_supported(bool *supportedp) {
...
+        r = build_service_query(&qstr, security_label, registry->bustype, name);
+        if (r < 0)
+                return error_origin(r);
+
+        r = aa_query_label(AA_DBUS_BIND, qstr, r, &allow, &audit);
+        if (r < 0)
+                return error_origin(-errno);
+
+        if (audit)
+                bus_apparmor_log("apparmor=\"%s\" operation=\"dbus_bind\" bus=\"%s\" name=\"%s\"", allowstr(allow), registry->bustype, name);
```
dvdhrm, line 423: `bustype` can contain quotes `"`, right? At least if we take it unmodified from the configuration. Probably doesn't matter much, though.

src/util/apparmor.h
```diff
@@ -6,5 +6,29 @@
 
 #include <stdlib.h>
 
+typedef struct NameSet NameSet;
+
+enum {
+        _APPARMOR_E_SUCCESS,
+
+        APPARMOR_E_DENIED,
```
dvdhrm, line 14: Can you prefix these with `BUS_APPARMOR_*`? The reason for the `BUS_*` prefix in selinux was that we do not want to conflict with selinux names, which sometimes start with `selinux_*`. I think the same is warranted for apparmor for consistency, even if the library uses `aa_*` as prefix (otherwise we should really drop it from the functions as well).

dvdhrm (changes requested):
A bunch more comments inline. It looks good! I think we can mostly take it that way, just some minor corrections needed.

Thanks a lot! I think I now got a full overview of the thing, and I am generally not opposed at all to merging this.

src/bus/policy.c
sre: After removing bustype parsing NULL is a valid configuration IIUIC, so I suppose always calling `bus_apparmor_set_bus_type()` makes sense and then I need to also store the `bool apparmor` in the registry, so that I can add `if (!apparmor_enabled) return success;` to the check functions.

dvdhrm: This sounds fine to me. I assume NULL would be a missing <bustype> declaration? I'd be fine using an empty string then, I don't see why we would need to distinguish it from a missing declaration?

sre: Yes, NULL would be a missing declaration. Rewriting that into an empty string should be ok and then NULL could indeed be used to check if AA is disabled.

src/util/apparmor.c
```diff
@@ -93,3 +105,446 @@ int bus_apparmor_dbus_supported(bool *supportedp) {
...
+        c_assert(!registry->bustype);
+
+        registry->bustype = strdup(bustype);
+        if (!registry->bustype)
+                return error_origin(-ENOMEM);
+        return 0;
+}
+
+static bool is_unconfined(const char *context) {
+        return !strcmp(context, "unconfined");
```
dvdhrm, line 189: dbus-daemon checks for the mode to be equal to `unconfined`, or unset and the label equal to `unconfined`. Any particular reason this is different?

src/util/apparmor.c
```diff
@@ -93,3 +105,446 @@ int bus_apparmor_dbus_supported(bool *supportedp) {
...
+                         * on non-primary name owners.
+                         */
+                        c_rbtree_for_each_entry(ownership, &nameset->owner->ownership_tree, owner_node) {
+                                r = apparmor_message_query_name(check_send, security_label,
+                                                                bustype, receiver_context,
+                                                                ownership->name->name, path,
+                                                                interface, method, allow, &audit_tmp);
+                                if (r < 0)
+                                        return r;
+                                if (!allow)
```
dvdhrm, line 340: Shouldn't this be `if (!*allow)`?

src/util/apparmor.c
```diff
@@ -93,3 +105,446 @@ int bus_apparmor_dbus_supported(bool *supportedp) {
...
+                                        return r;
+                                if (audit_tmp)
+                                        *audit = 1;
```
dvdhrm, line 343: Shouldn't this conditional be before checking `allow`?

src/util/apparmor.c
```diff
@@ -93,3 +105,446 @@ int bus_apparmor_dbus_supported(bool *supportedp) {
...
+        _c_cleanup_(c_freep) char *condup = NULL;
+        char *security_label, *security_mode;
+        _c_cleanup_(c_freep) char *qstr = NULL;
+        int r;
+        /* the AppArmor API uses pointers to int for pointers to boolean */
+        int allow = false, audit = true;
+
+        if (!is_apparmor_enabled(registry))
+                return 0;
+
```
dvdhrm, line 400: Why does this function not check against `unconfined` like the others and dbus-daemon do?

src/util/apparmor.c
```diff
@@ -93,3 +105,446 @@ int bus_apparmor_dbus_supported(bool *supportedp) {
...
+ * denied.
+ *
+ * In case multiple names are available all are being checked and the function
+ * will deny access if any of them is denied by AppArmor.
+ *
+ * Return: 0 if the transaction is allowed, BUS_APPARMOR_E_DENIED if it is not,
+ *         or a negative error code on failure.
+ */
+int bus_apparmor_check_xmit(BusAppArmorRegistry *registry,
+                            bool check_send,
```
dvdhrm, line 455: dbus-daemon only ever checks on `send`, why do we have two calls one for send and one for receive?

Maybe this is a confusion where dbus-daemon checks both for sending and receival in a single call, but that is not the case. Checking at send and receive is different to checking the receiver during send. I think this should be `bus_apparmor_check_send()` just like selinux does.

dvdhrm (changes requested):
I now pushed this with a bunch of fixups on top. This still lacks the last commit (handling apparmor in the launcher, though). Note that I am not really sure that this currently matches the behavior of dbus-daemon. I will try to align it with dbus-daemon later. In the meantime, if you have any comments on the code I pushed, let me know. I tried to implement everything that I mentioned inline in this PR.

dvdhrm:
I am quite confused now where you got the transmission checks from. dbus-daemon simply checks with the destination-string and source-unique-id, but your code iterates the set of names on the destination. This will very easily break existing AppArmor rules. I also don't understand why that change was done. Is there any reason not to imitate what dbus-daemon does?

That is, I would make apparmor check whether the sender is allowed to send to the destination used in the message. Then check whether the receiver is allowed to receive from the unique-id of the sender. If either is false, refuse the transmission.

I can implement this, but I want to make sure you agree on that, and that this is indeed the intended behavior?

src/launch/launcher.c
```diff
@@ -1155,6 +1155,55 @@ static int launcher_set_policy(Launcher *launcher, Policy *policy, uint32_t *sys
...
+                        *apparmor_mode = CONFIG_APPARMOR_DISABLED;
+                } else if (enabled) {
+                        *apparmor_mode = CONFIG_APPARMOR_ENABLED;
+                } else {
+                        *apparmor_mode = CONFIG_APPARMOR_DISABLED;
+                }
+        } else if (*apparmor_mode == CONFIG_APPARMOR_REQUIRED) {
+                if (!enabled || !supported) {
+                        fprintf(stderr, "AppArmor required, but not supported. Exiting.\n");
+                        *apparmor_mode = CONFIG_APPARMOR_REQUIRED;
```
setharnold, line 1197: I expected to see a negative return value here, or perhaps an `exit(3)` or `_exit(2)` call.

src/bus/policy.c
```diff
@@ -835,6 +846,14 @@ int policy_snapshot_check_own(PolicySnapshot *snapshot, const char *name_str) {
         size_t i;
         int v, r;
 
+        r = bus_apparmor_check_own(snapshot->apparmor, snapshot->seclabel, name_str);
+        if (r) {
+                if (r == BUS_APPARMOR_E_DENIED)
+                        return POLICY_E_APPARMOR_ACCESS_DENIED;
+
```
setharnold, line 853: Are there other return values that mean "don't grant access"? eg. "internal error" or "silently deny" or similar? This currently feels like a 'fail open' way to write this check. (And two more later.)

src/util/apparmor.c
```diff
@@ -93,3 +105,446 @@ int bus_apparmor_dbus_supported(bool *supportedp) {
...
+        r = aa_query_label(AA_DBUS_BIND, qstr, r, &allow, &audit);
+        if (r < 0)
+                return error_origin(-errno);
+
+        if (audit)
+                bus_apparmor_log("apparmor=\"%s\" operation=\"dbus_bind\" bus=\"%s\" name=\"%s\"", allowstr(allow), registry->bustype, name);
+
+        if (is_complain(security_mode))
+                allow = 1;
+
```
setharnold, line 420: Should this `allow = 1` block be moved above the `bus_apparmor_log()` call, so that the `allowstr(allow)` generates an output that matches the decision?

src/util/apparmor.c
```diff
@@ -93,3 +105,446 @@ int bus_apparmor_dbus_supported(bool *supportedp) {
...
+        r = aa_query_label(AA_DBUS_EAVESDROP, qstr, r, &allow, &audit);
+        if (r < 0)
+                return error_origin(-errno);
+
+        if (audit)
+                bus_apparmor_log("apparmor=\"%s\" operation=\"dbus_eavesdrop\" bus=\"%s\" label=\"%s\"", allowstr(allow), registry->bustype, context);
+
+        if (is_complain(security_mode))
+                allow = 1;
```
setharnold, line 544: A similar question here, should this be moved above the logging line, so the string matches the result?

setharnold (commented):
Hello, a few small thoughts; it's more of a quick skim than in-depth review, and while I'm alright with AppArmor, I know next to nothing about dbus-broker.

Thanks

src/bus/policy.c
dvdhrm: This is a translation of error codes from one domain to another. If we happen to forget converting one, the result is the following `error_fold()`, which takes any unhandled error and turns it into a fatal error code. Hence, the worst that can happen is that we end up exiting, rather than just refusing the message-transaction. Not good, but also we do not accidentally fail the security check.

src/util/apparmor.c
dvdhrm: Sounds good!

setharnold: Thanks!

src/util/apparmor.c
```diff
@@ -535,46 +533,69 @@ int bus_apparmor_check_xmit(BusAppArmorRegistry *registry,
...
-                        allow ? "ALLOWED" : "DENIED",
-                        check_send ? "send" : "receive",
+        if (string_equal(sender_security_mode, "complain"))
+                src_allow = 1;
+        if (string_equal(receiver_security_mode, "complain"))
+                dst_allow = 1;
+
+        if (src_audit) {
+                bus_apparmor_log(registry,
+                        "apparmor=\"%s\" bus=\"%s\" path=\"%s\" interface=\"%s\" method=\"%s\" "
```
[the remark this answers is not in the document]
gegarcia: nevermind, "method" and "member" are both valid, so we need to add support for "method" on AppArmor log parsing tools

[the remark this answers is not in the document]
gegarcia: I created a [PR](https://gitlab.com/apparmor/apparmor/-/merge_requests/958) in AppArmor so we can parse method the same way as member so no change is needed here :)

src/util/apparmor.c
```diff
@@ -535,46 +533,69 @@ int bus_apparmor_check_xmit(BusAppArmorRegistry *registry,
...
+                        "label=\"%s\" peer_label=\"%s\"",
```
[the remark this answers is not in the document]
gegarcia: Ah, I also noticed that `operation="dbus_method_call"` is missing here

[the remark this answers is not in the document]
sre: I added mask and operation. Logging name is non-trivial, since dbus-broker has a NameSet possibly containing multiple names.

gegarcia:
From an AppArmor perspective, this LGTM. We updated our [regression tests upstream](https://gitlab.com/apparmor/apparmor/-/merge_requests/965) and they are passing with these patches.

src/launch/launcher.c
eslerm: exit is handled outside of this function on line 1250

perhaps the fprintf on line 1196 should move between lines 1250 and 1251 ?

eslerm (approved):
Hi @dvdhrm o/

@gegarcia and @setharnold from AppArmor have ACKd this PR

there is one outstanding review request that might be resolved by moving a line

evverx:
As far as I can tell the CI is currently unaware of AppArmor in the sense that `-Dapparmor=true` isn't passed to `meson` anywhere. At least CodeQL analyzes the apparmor fallback only as far as I can see because `libapparmor-dev` isn't installed there and the "autobuild" job can't figure out what to do. The other CI workflow passes `-Daudit=true` only. There it would probably make sense to build `dbus-broker` with both `-Dapparmor=true` and `-Dapparmor=false` to make sure it compiles fine with `-Werror` with and without libapparmor.

(I'm not sure it should be fixed here but I think it should be fixed one way or another).

Maryse47:
Arch kernels don't have apparmor dbus mediation patches so included so this is irrelevant.

dvdhrm:
The required AppArmor features to get D-Bus message security policies working are not part of upstream linux kernels. Basic AppArmor support has been included in upstream kernels, though.

So to get AppArmor working with `dbus-broker`, you would definitely need downstream patches. Ubuntu applies them to all their kernel builds.

dvdhrm (approved):
I applied some minor changes on top. This looks good. Thanks a lot!

dvdhrm:
Merged via #410. Thanks for everyone helping with this! With the upcoming linux-6.17 release, this should work without any downstream patches (assuming a recent apparmor user-space that adapts to the new kernel ABI).

PR #296, opened 2022-07-10
build: require c-ini and c-shquote only when launcher is enabled

jirutka:
These dependencies are used only in launcher.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -22,10 +22,8 @@ mod_pkgconfig = import('pkgconfig')
 #
 
 dep_cdvar = dependency('libcdvar-1')
-dep_cini = dependency('libcini-1')
 dep_clist = dependency('libclist-3')
 dep_crbtree = dependency('libcrbtree-3')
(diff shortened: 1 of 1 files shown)


dvdhrm:
Is there a particular reason you want to guard those dependencies? We usually consider the c-util dependencies part of the project, so we don't guard their inclusion, but only their linkage into the binary.

Generally, in dbus-broker we try to avoid any conditional compilation and instead compile everything for better compile-coverage, less conditionals during compilation, better testing, and vastly reduced test-matrix. The selinux and audit dependencies are special, as we do not control the recursively included deps there. Furthermore, some distros do not even build those libraries, so we really have to guard them.

I am not strictly against the compile-guards, but just wonder what the reason for that is? Note that we are not aware of anyone building dbus-broker without the launcher, anyway (besides testing), so this would not have a real effect, anyway, right?

dvdhrm:
Closing for now. If there are any further comments on my explanation, please feel free to resubmit this!

PR #299, opened 2022-09-15
Add missing log_append_here() when logging "Ready"

adamel:
Signed-off-by: Marcus Sundberg <marcus.sundberg@aptilo.com>


dvdhrm (approved):
Nice catch! Thanks!

PR #300, opened 2022-09-26
Drop '#include <libaudit.h>' from src/util/apparmor.c

floppym:
This header does not seem to be required.
Tested with -Dapparmor=true and -Daudit=true/-Daudit=false.

Bug: https://bugs.gentoo.org/873082


dvdhrm:
Looks good! Thanks!

Issue #302, opened 2022-11-29
Transient Unit name for applications


swick:
Dbus activated processes currently live in transient units with the scheme `dbus-<unique>-<ServiceName>@<instance>.service`. The [systemd Desktop Environments Integration specification](https://systemd.io/DESKTOP_ENVIRONMENTS/) says that applications should be in service or scopes which follow the following scheme `app[-<launcher>]-<ApplicationID>[@<RANDOM>].service`.

xdg-desktop-portal relies on this convention to derive the app id of a calling process (when the application is not authenticated by a sandboxing specific code path). This however breaks for apps started by dbus-broker. It would be nice if it followed the convention.

Can we change the schema to `app-dbus-<ServiceName>@<unique><instance>.service`? Not all services are applications. Should we change the schema only if the service has a corresponding desktop file?

dvdhrm:
Sounds good! I created a PR to implement this. I added a dash between `<unique>` and `<instance>`, but didn't check whether systemd actually allows this in the template argument. Anyway, please let me know if that PR looks good!

Btw., we should document that this becomes API and should not be changed in the future.

smcv:
Am I right in thinking that for services with `SystemdService=foo.service`, dbus-broker will start the systemd service named `foo.service` (the same as dbus-daemon does), with transient units only used for services that lack a `SystemdService` (as dbus-broker's equivalent of dbus-daemon's "traditional activation" in a child process)?

If that's true, then xdg-desktop-portal will not always be able to rely on this convention anyway, both for dbus-broker and for dbus-daemon, because `foo.service` doesn't indicate an app ID.

dvdhrm:
Correct. I would mostly see this as an effort to make the naming scheme of dbus-broker API, and then stick to the spec as closely as possible.

smcv:
> I would mostly see this as an effort to make the naming scheme of dbus-broker API

Wouldn't that mean that adding a `SystemdService=fooapp.service` to an existing session service would be a functional regression on dbus-broker systems, because consumers of this "API" would no longer be able to identify its app-ID?

Something like this:

* start with a D-Bus `.service` file with `Name=com.example.FooApp`
    * dbus-broker starts it as a transient service, xdg-desktop-portal can identify it as part of `com.example.FooApp` (which is a nice thing to have)
    * dbus-daemon starts it via traditional activation, xdg-desktop-portal can't identify it (not ideal but not a regression)
* now add `SystemdService=fooapp.service`
[...]

PR #303, opened 2022-11-29
launch/service: name transient units after desktop-spec

dvdhrm:
The experimental desktop-integration specification [1] proposes a nameing scheme for the units of desktop applications. Try to implement this scheme.

Cc: @swick #302 

[1] https://systemd.io/DESKTOP_ENVIRONMENTS/


teg (approved):

swick:
I'm still not sure if we should change this only for activated processes which are applications. If so, how would we classify applications (e.g. look for a corresponding .desktop file)?

PR #307, opened 2022-12-15
bus/driver: rework EPERM handling

dvdhrm:
A set of 3 patches reworking how we currently return EPERM from stubs. The first two commits are fixed that should go in anyway, the third is a suggestion on how to improve the situation with stubbed driver calls.

Cc: @draconicfae @teg


teg (approved):
Makes sense to me!

PR #308, opened 2022-12-29
launch: delay service-start until matches are installed

### LINKED ISSUE #304: Timeout on service calls to a systemd unit which fails
A service call to a systemd D-Bus activated service which fails usually returns immediately with an error along these lines:

```
Call failed: Could not activate remote peer: unit failed.
```

We found that in a real world scenario, sometimes the method call instead times out. 

```
Call failed: Connection timed out
```

It seems once dbus-broker is in that state, all service call to that service time out. Restarting the dbus-broker brings back the previous (expected) behavior. It seems in that state systemd is no longer attempting to start the service even. 

In our case, we use [dbus-fast](https://github.com/Bluetooth-Devices/dbus-fast) to talk to the Bluetooth daemon. The Bluetooth daemon does not get started due to a condition (`/sys/class/bluetooth` missing).  It seems that a race condition gets triggered quite easily in our setup (restarting our application Home Assistant Core triggers the problem about every other restart).
 
I've also successfully reproduced the problem using three instances of `busctl` while running `busctl monitor`. This was on a ODROID-N2 ARM board:

```
while [ true ]; do busctl call org.bluez / org.freedesktop.DBus.ObjectManager GetManagedObjects; done
```

It seems a race condition within dbus-broker causes all future calls to the method to timeout instead of returning an error.

dvdhrm:
This series delays the service-start until we correctly installed all matches. With #285 we now wait for `LoadUnit()` to return before installing our D-Bus matches, so we know the proper service object-paths. However, this change caused a race-condition where the service might spawn and fail faster than we can process the reply to `LoadUnit()`. This is quite unlikely, given that the `LoadUnit()` operation is scheduled first. However, with the coalesced message-transmissions of dbus-broker, the different messages we schedule are often transmitted in a single syscall and thus on slower machines this race is actually possible to hit.

Regardless, this needs to be fixed, so we simply delay the start-unit operation until the matches are installed. This series starts with a bunch of cleanups and then moves code around to make the series easier to review.

Additionally, given that we now delay access to `service->[argc,argv,unit]`, this series also properly caches the service-information and ensures we do not access changing data.

Possibly fixes #304.

### DIFF
--- a/src/launch/service.c
+++ b/src/launch/service.c
@@ -612,10 +612,6 @@ static int service_start_unit(Service *service) {
         _c_cleanup_(sd_bus_message_unrefp) sd_bus_message *method_call = NULL;
         int r;
 
-        r = service_watch_jobs(service);
-        if (r)
-                return error_trace(r);
-
         r = service_watch_unit(service, service->unit);
         if (r)
                 return error_trace(r);
@@ -645,10 +641,6 @@ static int service_start_transient_unit(Service *service) {
         _c_cleanup_(c_freep) char *unit = NULL, *escaped_name = NULL;
         const char *unique_name;
         int r;
-
-        r = service_watch_jobs(service);
-        if (r)
-                return error_trace(r);
 
         r = sd_bus_get_unique_name(launcher->bus_regular, &unique_name);
         if (r < 0)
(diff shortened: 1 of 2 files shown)


teg (approved):
I did not test this, but changes look good!

agners:
I've tested this on the affected system using the commands  mentioned in #304 as well as our actual application. In both cases the race-condition no longer occurred, every call ended with the expected response (`Could not activate remote peer: unit failed.`).

teg:
Thanks @agners and @dvdhrm !

teg:
@dvdhrm merge failed, might need a manual rebase?

dvdhrm:
(rebased on main)

Issue #315, opened 2023-06-02
dbus-broker blocks method calls even if selinux is in permissive mode


DaanDeMeyer:
I'm trying to enable selinux in permissive mode in systemd's mkosi development environment (kernel cmdline: selinux=1 enforcing=0). This leads to an unbootable system on Fedora 38 where I see many log messages such as the following:

```
[   14.082895] dbus-broker[537]: A security policy denied :1.11 to send method call /org/freedesktop/DBus:org.freedesktop.DBus.Hello to org.freedesktop.DBus.
[   14.089441] dbus-broker[537]: A security policy denied :1.11 to send method call /org/freedesktop/DBus:org.freedesktop.DBus.AddMatch to org.freedesktop.DBus.
[   14.091191] dbus-broker[537]: A security policy denied :1.11 to send method call /org/freedesktop/DBus:org.freedesktop.DBus.GetNameOwner to org.freedesktop.DBus.
[   14.093315] dbus-broker[537]: A security policy denied :1.11 to send method call /org/freedesktop/DBus:org.freedesktop.DBus.AddMatch to org.freedesktop.DBus.
[   14.096283] dbus-broker[537]: A security policy denied :1.11 to send method call /org/freedesktop/DBus:org.freedesktop.DBus.GetNameOwner to org.freedesktop.DBus.
[   14.104926] dbus-broker[537]: A security policy denied :1.11 to send method call /org/freedesktop/DBus:org.freedesktop.DBus.AddMatch to org.freedesktop.DBus.
[   14.108630] dbus-broker[537]: A security policy denied :1.11 to send method call /org/freedesktop/DBus:org.freedesktop.DBus.GetNameOwner to org.freedesktop.DBus.
[   14.111382] dbus-broker[537]: A security policy denied :1.11 to send method call /org/freedesktop/DBus:org.freedesktop.DBus.AddMatch to org.freedesktop.DBus.
[   14.114287] dbus-broker[537]: A security policy denied :1.11 to send method call /org/freedesktop/DBus:org.freedesktop.DBus.GetNameOwner to org.freedesktop.DBus.
[   14.119286] dbus-broker[537]: A security policy denied :1.11 to send method call /org/freedesktop/DBus:org.freedesktop.DBus.AddMatch to org.freedesktop.DBus.
[...]

PR #318, opened 2023-06-12
selinux: support permissive mode

dvdhrm:
Add support for selinux permissive mode. Since we use `selinux_check_access()` rather than the deprecated direct AVC queries (which dbus-daemon uses), we don't get the automatic enforcing-mode check of the AVC helpers. Instead, we have to manually query `security_getenforce()` to get the value.

This implementation is modeled similar to the implementation in systemd (src/core/selinux-access.c). It uses a similar optimization via the status-page to avoid an additional syscall on each policy check.

Reported in #315.

### DIFF
--- a/src/util/selinux-fallback.c
+++ b/src/util/selinux-fallback.c
@@ -16,6 +16,10 @@ bool bus_selinux_is_enabled(void) {
         return false;
 }
 
+bool bus_selinux_is_enforcing(void) {
+        return false;
+}
+
 const char *bus_selinux_policy_root(void) {
         return NULL;
 }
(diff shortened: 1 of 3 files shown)


DaanDeMeyer (approved):
Any chance this could be backported to Fedora 38? (Not sure if you or another contributor also maintain that package)

teg (approved):

DaanDeMeyer:
Backported this feature to rawhide: https://src.fedoraproject.org/rpms/dbus-broker/pull-request/6

DaanDeMeyer:
cc @dvdhrm could you take a look at that PR for the dbus-broker rpm?

Issue #319, opened 2023-07-04
dbus-broker.service timeout when there is not enough entropy


agners:
On an Arm embedded platform (Rockchip RK3568 based) I came across the situation that starting of the `dbus-broker.service` timed out due to insufficient entropy (`/proc/sys/kernel/random/entropy_avail` was around 50 at the time of the service start, and did not fill up until 90s passed). I've been using libexpat 2.5.0, so the most recent version.

Reordering the service after `systemd-random-seed.service` helped to delay `dbus-broker.service` until enough entropy was available. It seems that is a recommended way of handling services which require randomness, from [systemd-random-seed.service documentation](https://www.freedesktop.org/software/systemd/man/systemd-random-seed.service.html):

> This new seed is retrieved synchronously from the kernel, which means the service will not complete start-up until the random pool is fully initialized. On entropy-starved systems this may take a while. This functionality is intended to be used as synchronization point for ordering services that require an initialized entropy pool to function securely (i.e. services that access /dev/urandom without any further precautions).

Should there be an `After=systemd-random-seed.service` in the `dbus-broker.service` file?

However, I've noticed that other services which require/order after `dbus.socket` (e.g. when using `BusName=`) still fail since they do not wait for `dbus-broker.service`... Not sure if/how that can be properly fixed, this is probably more a systemd issue.

dvdhrm:
Do you know where exactly it timed out? We use `AT_RANDOM` from auxval exclusively for this reason. This might be in some dependency like sd-bus or libselinux. Maybe we can avoid any getrandom() call.

I doubt that delaying dbus for >90s during boot would be an acceptable experience. So yeah, preferably, we just avoid any of those blocking calls.

dvdhrm:
From the linked bug-report this seems to be in libexpat. I think we could use `XML_SetHashSalt()` to provide `AT_RANDOM` as salt for libexpat, and thus avoid blocking on `/dev/urandom` reads.

agners:
> From the linked bug-report this seems to be in libexpat. I think we could use `XML_SetHashSalt()` to provide `AT_RANDOM` as salt for libexpat, and thus avoid blocking on `/dev/urandom` reads.

I am not 100% sure about libexpat being the culprit, but it seems very likely.

Thanks for looking into this! I'll test the patch you've created on the platform which is affected, then we know for sure if that was indeed the culprit!

agners:
Even with #320 applied, it seems that `dbus-broker-launch` still blocks:

[...]

PR #320, opened 2023-07-05
launcher/config: use AT_RANDOM for libexpat hash seeds

dvdhrm:
A small series with 2 minor fixes to the libexpat handling, and finally a workaround to forward AT_RANDOM to libexpat, rather than making libexpat fetch its own entropy from the kernel.

Cc: @agners #319

### DIFF
--- a/src/launch/config.c
+++ b/src/launch/config.c
@@ -1230,6 +1230,7 @@ void config_parser_init(ConfigParser *parser) {
         *parser = (ConfigParser)CONFIG_PARSER_NULL(*parser);
 
         parser->xml = XML_ParserCreate(NULL);
+        c_assert(parser->xml);
 }
 
 /**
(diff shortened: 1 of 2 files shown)


agners:
@dvdhrm the patch still lead to salt initialization in the parser. The reason is that `XML_ParserReset` clears the salt. We need to add a `XML_SetHashSalt` after `XML_ParserReset` as well.

I've hacked in a static salt there, and that indeed fixes the problem on my target board completely (as in, no more hang/read from `/dev/urandom`.

dvdhrm:
Oh, indeed, `XML_ParserReset()` calls `initParser()` and resets all the fields. I changed the patch to store the salt in the parser context and set it after resetting the parser.

agners (approved):
Tested on the problematic board here, no more blocking :tada: 

Thanks!

Issue #322, opened 2023-07-11
memfd_create() without MFD_EXEC nor MFD_NOEXEC_SEAL


gdamjan:
Recent kernels complain about dbus-broker-launch using memfd_create
```
memfd_create() without MFD_EXEC nor MFD_NOEXEC_SEAL, pid=281 'dbus-broker-lau'
```

systemd recently fixed this too with this PR
https://github.com/systemd/systemd/pull/26800

Would it be fine if we copy the same code/memfd wrapper from systemd to dbus-broker?

dvdhrm:
Thanks a lot for the hint. I tried to address the issue in #323, let me know if you think anything else is needed!

gdamjan:
looks good to me - as far as I can understand C code :)

dvdhrm:
Merged. Thanks a lot for the report!

PR #323, opened 2023-07-14
memfd: provide proper abstractions

dvdhrm:
Provide proper abstractions for memfd operations. This is especially necessary with the new `MFD_EXEC` and `MFD_NOEXEC_SEAL` features of the kernel, which requires all memfd users to update the code to use one of the new flags.

Unfortunately, they decided against providing `MFD_NOEXEC`, but instead just provide `MFD_NOEXEC_SEAL`, which is a combination of `MFD_ALLOW_SEALING`, `MFD_NOEXEC`, and `F_SEAL_EXEC`. This is really awkward to use in helpers, but I think I got something that works well enough.

Reported in #322.

### DIFF
--- a/src/launch/test-config.c
+++ b/src/launch/test-config.c
@@ -9,7 +9,7 @@
 #include "launch/config.h"
 #include "launch/nss-cache.h"
 #include "util/dirwatch.h"
-#include "util/syscall.h"
+#include "util/misc.h"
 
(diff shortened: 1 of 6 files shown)


dvdhrm:
(updated to use `MFD_EXEC` with journald since it rejects `F_SEAL_EXEC`)

PR #324, opened 2023-07-31
launcher: start unit immediately as unprivileged user

bluca:
It is better to never have privileges rather than start with them and remove them later, as the attack surface is reduced, and there are fewer things to do before being 'ready'. Nowadays systemd can run the service as the appropriate user/group out of the box.

When starting as root files in /proc/self/fdinfo/ will be owned as root and set to 400, so we cannot read them. Nowadays it is not necessary to start as root when running under systemd, so just add User/Group with the configured user to the system unit. Add a meson option to let users configure the user, and default to the same as dbus-daemon's default, 'messagebus'.

If libaudit support is enabled, add AmbientCapabilities=CAP_AUDIT_WRITE so that we can still write to the audit log.

Same change for dbus-daemon: https://gitlab.freedesktop.org/dbus/dbus/-/merge_requests/399

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -55,6 +55,9 @@ use_audit = get_option('audit')
 if use_audit
         dep_libaudit = dependency('audit', version: '>=3.0')
         dep_libcapng = dependency('libcap-ng', version: '>=0.6')
+        conf.set('ambientcaps', 'AmbientCapabilities=CAP_AUDIT_WRITE')
+else
+        conf.set('ambientcaps', '')
 endif
 
 #
--- a/src/launch/launcher.c
+++ b/src/launch/launcher.c
@@ -1086,6 +1086,14 @@ static int launcher_parse_config(Launcher *launcher, ConfigRoot **rootp, NSSCach
         /* Remember if our at_console compat logic is needed */
         launcher->at_console = at_console;
 
+        /*
+         * If we are not running as root but systemd launched us as a non-privileged user, enforce that
+         * the configuration is coherent with the unit setting and return an error otherwise, to avoid
+         * unexpected results.
+         */
+        if (launcher->uid != (uint32_t)-1 && geteuid() != 0 && geteuid() != launcher->uid)
+                return LAUNCHER_E_INVALID_CONFIG;
+
         return 0;
 }
 
--- a/src/units/system/dbus-broker.service.in
+++ b/src/units/system/dbus-broker.service.in
@@ -17,6 +17,9 @@ PrivateTmp=true
 PrivateDevices=true
 ExecStart=@bindir@/dbus-broker-launch --scope system --audit
 ExecReload=@bindir@/busctl call org.freedesktop.DBus /org/freedesktop/DBus org.freedesktop.DBus ReloadConfig
+User=messagebus
+Group=messagebus
+@ambientcaps@
 
 [Install]
 Alias=dbus.service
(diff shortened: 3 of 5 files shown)


dvdhrm:
I would actually prefer to always set the user and let distributions strip the lines from the files if they decide against it (is there precedent? Does anyone *not* want this?). Similarly, I would always include `CAP_AUDIT_WRITE`, but I am fine with your approach as well.

The only issue I have is that if the configuration contains a `user=` line, the daemon will fail changing to that user. And since we do not ship a dbus-configuration file, but use the reference configuration, this is a bit of an issue. Furthermore, are you sure this works with NSS? So far, the launcher has been running as root to ensure compatibility with existing NSS modules.

bluca:
> I would actually prefer to always set the user and let distributions strip the lines from the files if they decide against it (is there precedent? Does anyone _not_ want this?). Similarly, I would always include `CAP_AUDIT_WRITE`, but I am fine with your approach as well.

Yeah I think everyone uses the same, and if not adding a drop-in is trivial, so I've dropped the new meson option. I've left the ambient caps setting conditional on the build time audit option, as I think it's nice to add the additional permission only if it can actually be used, and omit it otherwise, least privilege principle and all that.

> The only issue I have is that if the configuration contains a `user=` line, the daemon will fail changing to that user. And since we do not ship a dbus-configuration file, but use the reference configuration, this is a bit of an issue.

I'd argue that's a configuration mistake - if the `user=` in the config file and the `User=` in the unit are not configured to be the same, then it's a packaging/build problem. The unit can be trivially adapted with a drop-in. There's immediately an error on startup, so it should be hard to miss.

> Furthermore, are you sure this works with NSS? So far, the launcher has been running as root to ensure compatibility with existing NSS modules.

What do you mean, specifically? I tested this in a Debian testing VM and didn't see any issue (including logging audit messages), but if there's something specific you want me to try I can do that

bluca:
FYI, corresponding change was merged in dbus-daemon: https://gitlab.freedesktop.org/dbus/dbus/-/merge_requests/399

teg:
I like this very much!

If the logic is that we respect the systemd unit and fail if the dbus config don't match, couldn't we check/fail early and delete a bunch of code that deals with dropping privileges?

bluca:
I can add a check as soon as the config is parsed yes. Regarding the other point, it depends on whether you want to retain the optional ability to start as root or not. If you don't, then I can remove it.

teg:
I don't see the point of keeping the ability to start as root if the service files we ship anyway won't do that. But maybe I'm missing something... I'll let @dvdhrm comment...

bluca:
I agree, the lesser code around to deal with privileges the better, but I'll leave the decision to you folks

bluca:
Yeah it uses setuid, but I very very much prefer to avoid that, it wouldn't be worth it.

But there's still one bit I am not following: the priv drop when system.conf sets the user to messagebus (ie, pretty much always) is done immediately on launcher startup before going into the main loop, and after fork/before exec of the broker. I can see both processes running as messagebus. How is it able to start units then?

dvdhrm:
I think one strategy to solve this is to add polkit-rules that allow the `dbus` user to run `StartTransientUnit()` and `StartUnit()`. I don't think we need any other privileges. However, being able to run `StartTransientUnit()` seems equivalent to root-privileges.

Yes, I see the advantage in running as non-root, even if we have enough privileges to spawn root-shells. But given that we drop privileges so early, this is very low-priority for me.

If someone wants to implement this, please go ahead! I am going to close this for now, since this implementation cannot be merged as it is. But If you want to pick this up, please feel free to re-open.

PR #353, opened 2024-04-11
test: follow-up on the integration tests

dvdhrm:
A collection of follow-ups on the recent integration test additions by @mrc0mmand. I think the major two changes are:

- I dropped the debug-statements that print the selected binary in each test-run. If we want this, lets introduce this properly and print it once per test, rather than on each use.
- Change `install-tests` to `tests` in meson_options.txt. See the commit-message for the rationale.


dvdhrm:
I will merge this now, since I have to prepare a bug-fix release, and I would like the configuration change in. But I will happily change things later on, even if this is merged, so please comment if anything needs further discussion!

mrc0mmand:
> I will merge this now, since I have to prepare a bug-fix release, and I would like the configuration change in. But I will happily change things later on, even if this is merged, so please comment if anything needs further discussion!

Looks good to me, thank you!

Maybe one small comment about [d630cd1](https://github.com/bus1/dbus-broker/pull/353/commits/d630cd12250290eb24487a5b097b3ed375865107) - printing this out multiple times per test is indeed annoying but having some confirmation that we run the tests against an expected binary would be nice. But that's definitely a minor detail that can be figured out later.

PR #358, opened 2024-04-16
test: run reference tests in CI

mrc0mmand:
Let's finish the reference tests saga and make use of installed reference tests ([0]) provided by a recently introduced dbus-broker-tests package ([1]).

This commit introduces a simple test that runs the reference tests (and benchmarks) against both dbus-daemon and dbus-broker.

[0] https://github.com/bus1/dbus-broker/pull/351
[1] https://src.fedoraproject.org/rpms/dbus-broker/pull-request/14


mrc0mmand:
Everything seems to be working fine. The only test combination that's currently skipped is `bench-message` with `dbus-daemon`, since it takes an _ungodly_ amount of time (over 13 minutes compared to just ~32 seconds with `dbus-broker`). Not sure if we want to slow down our CI with that.

dvdhrm:
`bench-message` installs up to half a million match rules and then measures message transactions. `dbus-daemon` iterates match-rules linearly, and might thus take a lot of time to perform this benchmark.

There might be another reason for it being slow, but this strikes me as the likely cause. We can also skip the test directly in bench-message.c, just like we do in test-fdstream.c.

mrc0mmand:
I thought about skipping `bench-message` directly, but there's nothing _wrong_ with - the benchmark work just _fine_, it's just horribly slow. But I really have no preference here, either way works just fine for me.

PR #359, opened 2024-04-17
test: run dbus-broker under ASan and UBsan

mrc0mmand:
Let's introduce a test that runs dbus-broker under Address Sanitizer and Undefined Behavior Sanitizer, while running other tests against it.

The setup to achieve this is slightly convoluted, since we need to run (and restart) sanitized dbus-broker without nuking the host machine. For that we setup an nspawn-container that re-uses host's rootfs (to some degree) and overlays our additions on top of that. This way we can test (not-only) the full user-space boot with sanitized dbus-broker without risking "damage" to the host machine.


mrc0mmand:
@evverx as [promised](https://github.com/bus1/dbus-broker/pull/348#discussion_r1537586526) here's a PoC of a test that runs dbus-broker under ASan and UBSan while we hammer it with other tests on the side. It got _slightly_ more involved, to protect the host machine, and LSan needed some extra care as well (see the comments in the test code).

It's different from the original idea of running existing tests on a sanitized build, since I currently have no clue how to incorporate this into the whole Packit/TestingFarm infra (and I'm also not quite sure how well it would handle dying dbus and collecting artifacts from such machines).

If I run the test in one of my Arch VMs against the latest upstream, it reports a leak after I run dfuzzer just on the D-Bus control interface, so it looks like it's doing something :)

```
# TMT_TEST_DATA=~/logs DBUS_BROKER_TREE=$PWD test/integration/fuzz/sanitizers/test.sh
...
[688762.734598] dbus-broker[66]: A security policy denied :1.16 to send method call /org/freedesktop/DBus:org.freedesktop.DBus.Peer.GetMachineId to org.freedesktop.DBus.
[688762.735575] dbus-broker[66]: A security policy denied :1.16 to send method call /org/freedesktop/DBus:org.freedesktop.DBus.Peer.Ping to org.freedesktop.DBus.
[688763.062724] dbus-broker[66]: Dispatched 1954 messages @ 198(±297)μs / message.
[688763.063133] systemd[1]: Stopping D-Bus System Message Bus...
[688763.105139] dbus-broker-launch[65]: =================================================================
[688763.105437] dbus-broker-launch[65]: ==65==ERROR: LeakSanitizer: detected memory leaks
[688763.105437] dbus-broker-launch[65]: Direct leak of 880 byte(s) in 10 object(s) allocated from:
[688763.105564] dbus-broker-launch[65]:     #0 0x75fe014e2cc1 in __interceptor_calloc /usr/src/debug/gcc/gcc/libsanitizer/asan/asan_malloc_linux.cpp:77
[688763.105564] dbus-broker-launch[65]:     #1 0x5b9d8d5fe8f7 in policy_record_new_xmit ../src/launch/policy.c:56
[688763.105564] dbus-broker-launch[65]:     #2 0x5b9d8d609364 in policy_import_send ../src/launch/policy.c:460
[688763.105564] dbus-broker-launch[65]:     #3 0x5b9d8d60f44f in policy_import ../src/launch/policy.c:696
[688763.105564] dbus-broker-launch[65]:     #4 0x5b9d8d5c018b in launcher_load_policy ../src/launch/launcher.c:1096
[688763.105564] dbus-broker-launch[65]:     #5 0x5b9d8d5c1943 in launcher_reload_config ../src/launch/launcher.c:1195
[688763.105564] dbus-broker-launch[65]:     #6 0x5b9d8d5c240c in bus_method_reload_config ../src/launch/launcher.c:1296
[688763.105564] dbus-broker-launch[65]:     #7 0x75fe0131b45c  (/usr/lib/libsystemd.so.0+0x2f45c) (BuildId: 540b0b05a1f8668a797c3ff583cd0b19ff6962f4)
[688763.105564] dbus-broker-launch[65]:     #8 0x75fe01334b1a  (/usr/lib/libsystemd.so.0+0x48b1a) (BuildId: 540b0b05a1f8668a797c3ff583cd0b19ff6962f4)
[688763.105734] dbus-broker-launch[65]: Direct leak of 88 byte(s) in 1 object(s) allocated from:
[688763.105877] dbus-broker-launch[65]:     #0 0x75fe014e2cc1 in __interceptor_calloc /usr/src/debug/gcc/gcc/libsanitizer/asan/asan_malloc_linux.cpp:77
[688763.105877] dbus-broker-launch[65]:     #1 0x5b9d8d5fe8f7 in policy_record_new_xmit ../src/launch/policy.c:56
[688763.105877] dbus-broker-launch[65]:     #2 0x5b9d8d609364 in policy_import_send ../src/launch/policy.c:460
[688763.105877] dbus-broker-launch[65]:     #3 0x5b9d8d60f44f in policy_import ../src/launch/policy.c:696
[688763.105877] dbus-broker-launch[65]:     #4 0x5b9d8d5c018b in launcher_load_policy ../src/launch/launcher.c:1096
[688763.105877] dbus-broker-launch[65]:     #5 0x5b9d8d5c313f in launcher_run ../src/launch/launcher.c:1347
[688763.105877] dbus-broker-launch[65]:     #6 0x5b9d8d5b270c in run ../src/launch/main.c:151
[688763.105877] dbus-broker-launch[65]:     #7 0x5b9d8d5b2949 in main ../src/launch/main.c:174
[688763.105877] dbus-broker-launch[65]:     #8 0x75fe00b1cccf  (/usr/lib/libc.so.6+0x25ccf) (BuildId: c0caa0b7709d3369ee575fcd7d7d0b0fc48733af)
[688763.106017] dbus-broker-launch[65]: SUMMARY: AddressSanitizer: 968 byte(s) leaked in 11 allocation(s).
[688763.115803] systemd[1]: dbus-broker.service: Main process exited, code=exited, status=1/FAILURE
[688763.115956] systemd[1]: dbus-broker.service: Failed with result 'exit-code'.
[688763.116408] systemd[1]: Stopped D-Bus System Message Bus.
[688763.116595] systemd[1]: dbus-broker.service: Consumed 2.112s CPU time, 142.2M memory peak, 0B memory swap peak
```

But it doesn't appear to ... appear in CI, interesting.

evverx:
> If I run the test in one of my Arch VMs against the latest upstream, it reports a leak after I run dfuzzer just on the D-Bus control interface, so it looks like it's doing something. But it doesn't appear to ... appear in CI, interesting

I'm guessing the policy triggering it isn't included in the Fedora base image used by the CI. It should probably be possible to track that policy down by removing the policies on the Arch VM one by one.

test/integration/fuzz/sanitizers/test.sh
```diff
@@ -0,0 +1,207 @@
...
+# So, this one is a _bit_ convoluted. We want to run dbus-broker under sanitizers, but this bears a couple of
+# issues:
+#
+#   1) We need to restart dbus-broker (and hence the machine we're currently running on)
+#   2) If dbus-broker crashes due to ASan/UBSan error, the whole machine is hosed
+#
+# To make the test a bit more robust without too much effort, let's use systemd-nspawn to run an ephemeral
+# container on top of the current rootfs. To get the "sanitized" dbus-broker into that container, we need to
+# prepare a special rootfs with just the sanitized dbus-broker (and a couple of other things) which we then
+# simply overlay on top of the ephemeral rootfs in the container.
```
evverx, line 89: I can't say I fully understand what's going on here :-) but assuming the next step would be to run it all under Valgrind too I wonder if it's possible to somehow split the script and move the test suite itself to a separate bash script that could be run in two different places?

mrc0mmand: It _should_ be relatively easy to factor the container shenanigans out into a separate helper, I'll check what's the actual reality :)

evverx: I have one more question. Is it run on VMs/bare-metal machines where coredumps can be collected? I'm not sure what should happen when dbus-broker crashes/segfaults or whatever.

mrc0mmand: Ah, good point. Until recently systemd-coredump would leave coredumps from containers on the host, but since systemd/systemd@a108c43e36d3ceb6e34efe37c014fc2cda856000 it forwads the coredumps to the container to handle, so this might need some extra attention as well. We could either disable the forwarding (`CoredumpReceive=no` in the respective `systemd-nspawn@.service` instance) and just use `coredumpctl` on the host, configure `systemd-journald` to store the coredumps in the journal (since the container journal is already exported on the host), or bind mount a directory from the host to `/var/lib/systemd/coredump` so the coredumps don't vanish together with the container.

I like the first option the most (and since we use host's /usr we shouldn't have any issues with mismatches symbols), but I'll play with this a bit to see how well it works in practice.

mrc0mmand: (Apologies for dropping the ball on this, got very busy with some internal stuff. Hopefully I'll get back to this soon.)

mrc0mmand:
@evverx I factored out the common parts into a separate utility script, prepped another test that runs dbus-broker under Valgrind, and it seems to work (or at least Valgrind seem to [complain a lot](https://artifacts.dev.testing-farm.io/8ac35342-6813-41da-94d4-7f6904d20761/work-upstream_cia63f69yj/plans/upstream_ci/execute/data/guest/default-0/fuzz/valgrind-3/output.txt)). However, it will need a bit more polish I'll move the last two commits into a separate branch with the next push (which might take a bit, as I'll like to gather some coverage reports for the _sanitized_ dbus-broker to see how we could improve the ASan+UBSan test), so it doesn't block the sanitizer test.

test/integration/fuzz/sanitizers/main.fmf
```diff
@@ -0,0 +1,20 @@
+summary: Run dbus-broker under sanitizers
+test: ./test.sh
+require:
+    - dbus-broker
+    - dbus-daemon
+    - dfuzzer
+    - expat-devel
+    - gcc
```
evverx, line 8: I think it would be better to build it with clang because it comes with more checks. (Ideally it would be great to build it with both but if it's either gcc or clang I'd pick clang :-))

test/integration/fuzz/sanitizers/test.sh
```diff
@@ -0,0 +1,187 @@
...
+meson setup "$BUILD_DIR" --wipe -Db_sanitize=address,undefined -Dprefix=/usr
+ninja -C "$BUILD_DIR"
+meson test -C "$BUILD_DIR" --timeout-multiplier=2 --print-errorlogs
+
+: "=== Run tests against dbus-broker running under sanitizers ==="
+# So, this one is a _bit_ convoluted. We want to run dbus-broker under sanitizers, but this bears a couple of
+# issues:
+#
+#   1) We need to restart dbus-broker (and hence the machine we're currently running on)
+#   2) If dbus-broker crashes due to ASan/UBSan error, the whole machine is hosed
```
evverx, line 72: As far as I understand it should be applicable to "plain" builds too in the sense that if dbus-broker crashes/gets stuck there the testbed is hosed as well. I wonder if this container trick should be used in the integration tests generally? (Just to clarify I'm not saying it should be implemented here and I don't fully understand what's going on under the hood. I'm guessing it isn't possible to prevent the package from being installed and started)

test/integration/util.sh
```diff
@@ -0,0 +1,224 @@
...
+    #     we're interested in (in this case it's pretty simple, since dbus-broker installs everything under /usr,
+    #     and we need /etc with our dbus-broker.service override)
+    #   - since the whole container is ephemeral, use --link-journal=host, so the journal directory for the
+    #     container is created on the _host_ under /var/log/journal/<machine-id> and bind-mounted into the
+    #     container; that way we can fetch the container journal for debugging even if something goes horribly
+    #     wrong
+    mkdir -p "/run/systemd/system/systemd-nspawn@$CONTAINER_NAME.service.d"
+    cat >"/run/systemd/system/systemd-nspawn@$CONTAINER_NAME.service.d/override.conf" <<EOF
+    [Service]
+# We'll handle the coredumps on the host instead
```
evverx, line 57: It's really cool that it's possible to collect coredumps on the host there. GH Actions are run in containers so coredumps go to the underlying host actions have no access to. I guess it's convincing enough to switch to this testing infrastructure :-)

test/integration/fuzz/sanitizers/test.sh
mrc0mmand: I thought about that as well, but I'm not sure if it's worth the trouble. I opted into the container shenanigans just because both the ASan and Valgrind tests change how dbus-broker is invoked (and under which user), which has _slightly_ higher probability of going south (and some services don't like restarting dbus in general).

However, even if dbus dies the machine still remains (somewhat) usable (and this scenario should be very unlikely in "plain" tests). I plan on adding an "at exit" test/task that'll run after all tests (or maybe after _each_ test) and collects possible coredumps and other useful logs, so even if we manage to crash dbus-broker in a "plain" test, we should still have necessary dumps and logs available to debug it.

Issue #316, opened 2023-06-05
dbus-broker should probably be sent to Coverity Scan automatically


evverx:
It's mostly prompted by https://bugs.launchpad.net/ubuntu/+source/dbus-broker/+bug/2015538/comments/5.

>   - upstream also runs https://scan.coverity.com/projects/dbus-broker
>    - last scanned on June 18, 2022
>      - could use git action

and I agree with @eslerm here.
 
It should be possible to borrow the action from `systemd` or `dfuzzer` to run that cron job daily on GitHub but to get it to work a secret named `COVERITY_SCAN_TOKEN` should be added to the repository: https://docs.github.com/en/actions/security-guides/encrypted-secrets#creating-encrypted-secrets-for-a-repository. The token can be found at https://scan.coverity.com/projects/dbus-broker?tab=project_settings.

Since Ubuntu is used on GHActions the AppArmor stuff should start being analyzed automagically too. Currently it's missing there.

In theory `dbus-broker` can be integrated as is and false positives can be suppressed manually there but it would be better to help coverity a bit with, say, `c_assert` and add `#ifdef __COVERITY__`.

cc @mrc0mmand.

mrc0mmand:
Just linking the respective [coverity.yml](https://github.com/dbus-fuzzer/dfuzzer/blob/master/.github/workflows/coverity.yml) and [coverity.sh](https://github.com/dbus-fuzzer/dfuzzer/blob/master/.github/workflows/coverity.sh). IIRC I synced them between the dfuzzer and systemd repos so it should be more or less the same stuff, sans the build dependencies, project name, and the actual build command.

The [systemd](https://github.com/systemd/systemd/blob/6f814d4f5d6ac3847bfd75914ac8608722fb3021/tools/coverity.sh#L47) version also sets the email address for the scan notification to a pre-defined one (which is currently my email) contrary to the dfuzzer version which uses email address from the latest commit. I did that in systemd to make sure I'll get notified on any possible fails (and because I think the scan notification is usually useless for people not involved in whole Coverity stuff).

evverx:
> contrary to the dfuzzer version which uses email address from the latest commit

That would be yours too I believe :-)

I've just sent dbus-broker with https://github.com/bus1/dbus-broker/commit/af6bf48ba27408725d14378e4a11139106506f3b included and Coverity confirmed that it's gone :-)

Anyway the scripts are ready. I can keep running it manually from time to time until the secret is added to the repository.

PR #363, opened 2024-05-07
ci: submit dbus-broker builds to Coverity automatically

mrc0mmand:
Let's make a full use of Coverity and submit the builds for analysis automatically every midnight. We can't do that for every PR, since there are quite strict rate limits that limit how many builds we can submit per day and per week (see [0]).

The action (and the script) requires two environment variables to be set - $COVERITY_SCAN_TOKEN for authentication (can be found here [1]), and $COVERITY_SCAN_NOTIFICATION_EMAIL for sending the email notification when the build analysis is done. Originally this email used to be set to the email from the latest commit, but since the author of that commit might not even have permissions to see the Coverity report it's best to set it to one of the dbus-broker maintainers.

Resolves: #316

[0] https://scan.coverity.com/faq#frequency
[1] https://scan.coverity.com/projects/dbus-broker?tab=project_settings

---

As mentioned in the commit description (and the GH Actions file), there are two environment variables that need to be configured in the dbus-broker repo for this action to work properly. The configuration can be found under repo settings -> Secrets and variables -> Actions -> Repository secrets:

image

As for the notification email - this is really up to the maintainers to pick one (feel free to use mine, since that's already the case for other upstream repos). You'll get one email every day with the results of the analysis, and potentially a second one with description of each flaw that was detected if the analysis detected something new.

I gave the action spin in https://github.com/mrc0mmand/dbus-broker/pull/3 and it [seems to work](https://github.com/mrc0mmand/dbus-broker/actions/runs/8984090005/job/24675196756?pr=3) as expected.

/cc @evverx


.github/workflows/coverity.yml
```diff
@@ -0,0 +1,36 @@
+---
+# vi: ts=2 sw=2 et:
+
+name: Coverity
+
+on:
+  schedule:
+    # Run Coverity daily at midnight
+    - cron:  '0 0 * * *'
+  pull_request:
```
evverx, line 10: I'm not sure secrets are available when PRs are opened. If they were it would be possible for anyone to "steal" them by opening PRs and sending them somewhere.

mrc0mmand: Yeah, this is mostly for maintainers (at least I do this in other repos) - if you open a PR from a branch directly in the upstream repo the secrets are accessible, so it makes debugging or changing the configuration less of a PITA. But in this case the action will fail even after the secrets are set up, since it's opened from a fork.

evverx: > it makes debugging or changing the configuration less of a PITA

Agreed. I think one option would be to special-case that by checking where PRs come from and skip it if they come from forks or, maybe, add a comment somewhere. Then again it should make it easier to spot PRs that should be reopened from branches if it fails loudly.

.github/workflows/coverity.sh
```diff
@@ -0,0 +1,62 @@
...
+
+    tar czf "$results_archive" "$results_dir"
+
+    set +x # this is supposed to hide COVERITY_SCAN_TOKEN
+    echo -e "\033[33;1mUploading Coverity Scan Analysis results...\033[0m"
+    response=$(curl \
+               --silent --write-out "\n%{http_code}\n" \
+               --form project="$COVERITY_SCAN_PROJECT_NAME" \
+               --form token="${COVERITY_SCAN_TOKEN:?}" \
+               --form email="${COVERITY_SCAN_NOTIFICATION_EMAIL:?}" \
```
evverx, line 47: I borrowed this https://github.com/avahi/avahi/pull/600 :-) Though as far as I can remember thanks to that spammy approach some folks got subscribed to the systemd dashboard and helped to fix some issues :-)

PR #366, opened 2024-05-23
util/misc: support kernels where `MFD_NOEXEC_SEAL` does not enable sealing

pobrn:
I am trying to squash the kernel's peculiar behaviour of implicitly enabling sealing on `MFD_NOEXEC_SEAL`. `dbus-broker` is one of the affected programs, but as noted below, only the test suite would break, normal operations wouldn't. At least as far as I can tell.

---

Newer Linux kernels do not unset `F_SEAL_SEAL` when `MFD_NOEXEC_SEAL` is supplied to `memfd_create()`. Currently, `misc_memfd()` has a workaround for this peculiarity of the kernel, however, this workaround will break.

Avoid the breakage by checking if the kernel added `F_SEAL_SEAL`, and only try to add it if it is not already present.

Note that this kernel uAPI change does not affect dbus-broker's normal operations since it never attempts to create a memfd with `MFD_NOEXEC_SEAL` and without `MFD_ALLOW_SEALING`. Only the test suite would be broken on newer kernels without this change.

---

v2: https://lore.kernel.org/linux-mm/20240524033933.135049-1-jeffxu@google.com/
v1: https://lore.kernel.org/linux-mm/20240513191544.94754-1-pobrn@protonmail.com/

### DIFF
--- a/src/util/misc.c
+++ b/src/util/misc.c
@@ -49,7 +49,7 @@
  *    Throw in `MFD_CLOEXEC` or `F_SEAL_SEAL` as required.
  *
  *  * If `MFD_NOEXEC_SEAL` is used without `MFD_ALLOW_SEALING`, sealing will
- *    be disabled (even though the kernel implicitly enables it).
+ *    be disabled (even though some kernel versions implicitly enable it).
  *
  *  * An initial set of seals is applied to the memfd, if specified in
  *    @seals. Note that this is not allowed if sealing was not enabled.
@@ -111,7 +111,7 @@ int misc_memfd(const char *name, unsigned int uflags, unsigned int useals) {
         }
 
         /*
-         * If we ended up passing `MFG_NOEXEC_SEAL` to the kernel, the kernel
+         * If we ended up passing `MFG_NOEXEC_SEAL` to the kernel, some kernel versions
          * will implicitly enable sealing. This is very unfortunate, so we
          * revert this if the caller did not explicitly allow it. To disable
          * sealing, simply set `F_SEAL_SEAL`, which is also what the kernel
@@ -126,9 +126,15 @@ int misc_memfd(const char *name, unsigned int uflags, unsigned int useals) {
          * into the kernel again.
          */
         if (seals) {
-                r = misc_memfd_add_seals(fd, seals);
-                if (r)
-                        return error_fold(r);
+                r = fcntl(fd, MISC_F_GET_SEALS);
+                if (r < 0)
+                        return error_origin(-errno);
+
+                if (seals & ~r) {
+                        r = misc_memfd_add_seals(fd, seals);
+                        if (r)
+                                return error_fold(r);
+                }
         }
 
         r = fd;


src/util/misc.c
```diff
@@ -49,7 +49,7 @@
  *    Throw in `MFD_CLOEXEC` or `F_SEAL_SEAL` as required.
  *
  *  * If `MFD_NOEXEC_SEAL` is used without `MFD_ALLOW_SEALING`, sealing will
- *    be disabled (even though the kernel implicitly enables it).
+ *    be disabled (even though older kernels implicitly enable it).
```
dvdhrm, line 52:
```suggestion
 *    be disabled (even though some kernel versions implicitly enable it).
```

src/util/misc.c
```diff
@@ -111,14 +111,24 @@ int misc_memfd(const char *name, unsigned int uflags, unsigned int useals) {
         }
 
         /*
-         * If we ended up passing `MFG_NOEXEC_SEAL` to the kernel, the kernel
+         * If we ended up passing `MFG_NOEXEC_SEAL` to the kernel, older kernels
```
dvdhrm, line 114:
```suggestion
         * If we ended up passing `MFG_NOEXEC_SEAL` to the kernel, some kernel versions
```

src/util/misc.c
```diff
@@ -111,14 +111,24 @@ int misc_memfd(const char *name, unsigned int uflags, unsigned int useals) {
...
+                if (r < 0)
+                        return error_origin(-errno);
+
+                /*
+                 * Newer kernels do not implicitly allow sealing on `MFG_NOEXEC_SEAL`,
+                 * so only try to set `F_SEAL_SEAL` if not already set by the kernel.
+                 */
+                if (!(r & MISC_F_SEAL_SEAL))
+                        seals |= MISC_F_SEAL_SEAL;
+        }
```
dvdhrm, line 131: How about you leave this part as it was and instead change the code below to this:

```
if (seals) {
        r = misc_memfd_get_seals(fd);
        if (r < 0)
                return error_fold(r);

        if (seals & ~r) {
                r = misc_memfd_add_seals(fd, seals);
                if (r)
                        return error_fold(r);
        }
}
```

This moves the `MISC_F_GET_SEALS` call down a bit, but safely works around the problem of the kernel not allowing F_SEAL_SEAL if F_SEAL_SEAL is already set.

dvdhrm (changes requested):
Thanks! I am certainly in favor of this change. I also replied on LKML. Few comments inline!

dvdhrm:
(I can also amend the changes myself when applying the patch. Let me know!)

src/util/misc.c
dvdhrm: Yeah, I know, but I think that is fine. It is now 3 instead of 2 syscalls in that particular path (i.e., creating a memfd with initial seals, which is nothing we do, anyway). If we decide to optimize this path, we can always create fast-paths later. I prefer if the default is the "works for all" solution.

dvdhrm:
I have not seen the force-push, sorry for missing that. I adjusted the PR to keep the 80ch comment line limit and added minor memfd adjustments as add-on commits. The roll-up PR is pending as #371, so I am closing this one.

Thanks a lot for pushing this forward! I still hope we can get the kernel side fixed.

PR #369, opened 2024-07-09
bus/match: fix matching against arguments beyond the zeroth

### LINKED ISSUE #367: Filtering signals by arguments does not work
Hi, for some reason i am not able to get signal filtering by argument value to work.

### Reproducer

1. start monitoring dbus with `# dbus-monitor --system "arg3='done'"`
2. try to submit systemd job with for example `# systemctl restart polkit`
3. check found signal

### Expected result
signal time=1716466590.572778 sender=:1.0 -> destination=(null destination) serial=5066 path=/org/freedesktop/systemd1; interface=org.freedesktop.systemd1.Manager; member=JobRemoved
   uint32 4996
   object path "/org/freedesktop/systemd1/job/4996"
   string "polkit.service"
   string "done"
### Actual result
None

Context:
I am developing daemon that is listening for JobRemoved signal of systemd and i would like to receive only signal related to a service i am watching. I am using dbus-python but so far i am not able to get filtering by argument to work.
Am i missing something or is it possible that there is a bug?

Environment i am using is Fedora Rawhide.
dbus-broker-36-2

Thanks for any help.

jbosboom:
Fixes #367.  I hit this too, also for JobRemoved signals, and @dvdhrm pinpointed the fix, so here it is.

I added both white-box and black-box tests.  The former are copies of the arg0 tests changed to match arg1.  The latter include negative tests (signals that should not be matched), which aren't present in the existing tests, but seem useful.  I didn't run the black-box tests against dbus-daemon because I don't see how.  I put the tests in the same commit as the fix based on looking at old commits in this repo, but they can be separate if you'd prefer.

@TomasKorbar I gave you Reported-by: under your Red Hat e-mail (based on recent commits in other repos).  Let me know if this should be under a personal e-mail instead or just omitted.

@dvdhrm You can add Co-developed-by: yourself if you want, as you did find the fix.

### DIFF
--- a/src/bus/match.c
+++ b/src/bus/match.c
@@ -422,12 +422,12 @@ static bool match_keys_match_metadata(MatchKeys *keys, MessageMetadata *metadata
                 return false;
 
         for (unsigned int i = 0; i < keys->filter.n_args || i < keys->filter.n_argpaths; i ++) {
-                if (keys->filter.args[i] && !(metadata->args[0].element == 's' && string_equal(keys->filter.args[i], metadata->args[i].value)))
+                if (keys->filter.args[i] && !(metadata->args[i].element == 's' && string_equal(keys->filter.args[i], metadata->args[i].value)))
                         return false;
 
                 if (keys->filter.argpaths[i]) {
                         if (!match_string_prefix(metadata->args[i].value, keys->filter.argpaths[i], '/', true) &&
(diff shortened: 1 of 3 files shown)


dvdhrm:
Thank you so much for moving this forward and supplying the tests on top! I included this in the pending roll-up PR #371, and once this lands I will prepare a release.

Again, thanks a lot to all of you!

PR #378, opened 2024-09-13
launcher: allow searching for system-services in /etc/ and /run/

bluca:
This is useful when an asset manager wants to install a system service while /usr/ is read-only (e.g.: local system services running on a different namespaced image)

spec and dbus-daemon corresponding change: https://gitlab.freedesktop.org/dbus/dbus/-/merge_requests/467

### DIFF
--- a/src/launch/launcher.c
+++ b/src/launch/launcher.c
@@ -925,6 +925,8 @@ static int launcher_load_standard_session_services(Launcher *launcher, NSSCache 
 
 static int launcher_load_standard_system_services(Launcher *launcher, NSSCache *nss_cache) {
         static const char *default_data_dirs[] = {
+                "/etc",
+                "/run",
                 "/usr/local/share",
                 "/usr/share",
                 "/lib",


dvdhrm:
Thanks! No objections from me!

Any consensus on whether to enable inotify-watches on the entry? And whether to enforce well-defined file-names for service entries there? I can wait merging these until they the discussion is settled. Personally, I agree with Smcv and we should not provide inotify-watches, but enforce well-defined file-names.

dvdhrm:
We enforce strict names on the system bus for security reasons, but warn only on the user bus.

I think we unconditionally watch all service dirs, though. I think the reasoning was to be consistent, not entirely sure TBH.

So I think the patch is fine as is.

bluca:
Ok no problem, I'll leave as-is then. We are adding /run/ as well in dbus-daemon, so added here as well.

bluca:
> https://gitlab.freedesktop.org/dbus/dbus/-/merge_requests/467

The dbus-daemon and spec side was merged just now

dvdhrm:
Thanks a lot!

Weird, I never noticed that /run isn't highest priority. So it is just a matter of time until we introduce /run/etc? :)

bluca:
...or /etc/run/ !!

Issue #145, opened 2018-06-18
[question] What is equivalent of 'dbus-run-session' command?


Rosanna4:
Currently KDE plasma wayland session is started as `Exec=dbus-run-session /usr/bin/startplasmacompositor` . How to change it to use dbus-broker instead?

dvdhrm:
There is currently no equivalent to `dbus-run-session`. The `dbus-broker` binary itself only implements the D-Bus Bus-logic, but not any surrounding environment. We provide `dbus-broker-launch` as a compatibility launcher for the user and system bus as described by the D-Bus spec. We do not provide launchers for other types of buses so far.

However, the launcher can be used for custom buses already (at-spi2 does that, for example). We want to be aware of these use-cases, though. So why exactly do you need `dbus-run-session`? Which distribution do you use? How do you start your graphical environment? Do you intentionally use a session bus rather than a user bus? Do you use any graphical greeter like gdm or kdm? Do you use systemd?

It is definitely possible to write something like dbus-run-session for dbus-broker, but I'd like not to, unless really necessary. So I first want to get a feeling why you use the tool, and maybe you can use some drop-in dbus-broker replacement already.

Rosanna4:
I'm using what [upstream currently provides](https://github.com/KDE/plasma-workspace/blob/master/plasmawayland.desktop.cmake#L2) so it isn't config or distro dependent. I noticed that starting plasma-wayland session script directly without dbus like:
`Exec=${CMAKE_INSTALL_FULL_BINDIR}/startplasmacompositor` works however I don't know if there are any hidden consequences and I can't answer why KDE guys are starting the script like this.

dvdhrm:
Perfect! Most distributions use systemd to start user and system buses these days. So the old session-bus-starter is obsolete. Back in the old days someone had to start the dbus session bus, so most desktop environments simply wrapped their start-up script in `dbus-run-session`. But if your distribution already manages the buses, it is even detrimental to use this, since you now might be unable to discover services that run on the other bus.

I am not sure what to do. I guess we should at least properly document this and recommend to everyone to drop the tool from their scripts.

jadahl:
What is the recommended way to run a nested dbus session, if not by using `dbus-run-session`? I mostly use it to run debug instances of GNOME Shell, as well as the for the CI pipeline when running inside Docker.

dvdhrm:
[...]

PR #380, opened 2024-10-23
RFC: Implement dbus-broker-session tool

nolange:
This is a functional replacement for dbus-run-session, as this is often a very useful tool for testing / isolation (#145)

Exit codes and behavior is very similar,
the implementation is mostly shared with dbus-broker-launch, adding some hooks to start and supervise the application.

The target is to provide a minimal implementation, which can be enabled by creating a symlink named  `dbus-broker-session` to `dbus-broker-launch`. Possibly this could be extended to also allow a symlink names `dbus-run-session`.
More complex functionality like supporting dbus-daemon or different `dbus-broker-launch` versions is not planned (like in  #321), but could be done with simple shell scripts.

I believe this to be the best approach, and if merged it will quickly be available in distributions.

Please give me some feedback if this is worth pursuing (more than a version that "works-for-me"). I know it needs another pass in regards of code-style atleast, so general stuff would be needed to be sorted out first.

-   prefer separate functions like `help` / `help_session` or common functions with flag like `parse_argv`
-   `posix_spawn` or `fork/exec`
-   error handling is ok? in regards to `sd_event` + the `error_fold` stuff
-   I used `getauxval` because its used already in dbus-broker. but I would prefer `getrandom`.?
-   socket path should try `RUNTIME_DIRECTORY`, `XDG_RUNTIME_DIR` first.

### DIFF
--- a/src/launch/main.c
+++ b/src/launch/main.c
@@ -2,13 +2,17 @@
  * D-Bus Broker Launch Main Entry
  */
 
+#include "launch/launcher.h"
+#include "util/error.h"
 #include <c-stdaux.h>
 #include <fcntl.h>
 #include <getopt.h>
 #include <stdlib.h>
+#include <spawn.h>
 #include <systemd/sd-daemon.h>
-#include "launch/launcher.h"
-#include "util/error.h"
+
+#include <sys/auxv.h>
+#include <sys/un.h>
 
 enum {
         _MAIN_SUCCESS,
@@ -21,6 +25,8 @@ static const char *     main_arg_configfile = NULL;
 static bool             main_arg_user_scope = false;
 static int              main_fd_listen = -1;
 
+#define SESSION_TOOL "dbus-broker-session"
+
 static void help(void) {
         printf("%s [GLOBALS...] ...\n\n"
                "Linux D-Bus Message Broker Launcher\n\n"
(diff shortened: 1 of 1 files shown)


nolange:
And I am not sure if [launcher_connect](https://github.com/bus1/dbus-broker/blob/be2875131bd717afc759e3bf5ba4547992ffdb9c/src/launch/launcher.c#L1247)
should connect to the "normal" user bus or the newly created one (which I currently do by setting `DBUS_SESSION_BUS_ADDRESS` early).

This is needed for authentication and further launching via systemd I suppose?
`dbus-run-session` can be called in an environment where the user dbus / socket is not available.
For example I am preparing a debian filesystem with an per-existing user, this needs calling `dconf`, `gsettings` and `gio` tools through `dbus-run-session` for the user **before** he is logged in for the first time.

Ideally `dbus-broker-session` could run without dbus-activation feature (I expect there to be problems launching via systemd anyway).

dvdhrm:
Thanks for the proposal! Two comments:

1) Can you explain why you share the process with the launcher rather than as a separate executable? I have to admit I think the merged approach makes this less readable than a clear separation.

2) Any reason not to simply pursue #321 instead? I mean #321 works and it is fully compatible to `dbus-run-session`. If you rip out the `clap` dependency and parse the cli manually, it should be good to go.

nolange:
I am concerned about how this new tool is going to end up in debian, buildroot and other linux systems.
The `dbus-broker-launch` tool already has pretty much everything thats needed, the tool in #321 is a parallel implementation needing a separate toolchain - it would pose trouble for example since I have embedded targets at work which don't have a toolchain with rust compiler.
And on debian you typically dont install via cargo but have to package every dependency, and rust doest seem to support DSOs which means you need to rebuild every time a dependency gets a bugfix.

I also asked about the logic in `launcher_connect`, if there are modifications needed then its alot easier if `dbus-run-session` is directly coupled to `dbus-broker-launch` (same project or same binary) vs. detecting the required behavior across fork/exec.

> 1. Can you explain why you share the process with the launcher rather than as a separate executable? I have to admit I think the merged approach makes this less readable than a clear separation.

It was easier to start that way, and I there are advantages for the "busybox" approach like getting the functionality deployed quickly without any packaging changes (debian policy means every binary needs a manpage for example).
Yeah, the readability is a downside.

PR #388, opened 2025-03-26
util/audit: Correct usage of audit logging functions

### LINKED ISSUE #387: util_audit_log should create valid audit records
The `audit_log_user_avc_message` function is used to log a user AVC message. However, this is being misused in https://github.com/bus1/dbus-broker/blob/main/src/util/audit.c#L104C5-L104C19 . When the type is `UTIL_AUDIT_TYPE_AVC`, it should use the current audit function. The other two cases should ideally use `audit_log_user_message`.

This has been pointed out in https://github.com/linux-audit/audit-userspace/pull/394 . The problem is that the `ausearch` utility expects some event types to match a certain argument order. For example the `USER_MAC_POLICY_LOAD`. It's not feasible to fix the parser issue within the ausearch parser due to performance optimization, for more details see the linked issue.

Cropi:
Update util_audit_log to use audit_log_user_message for UTIL_AUDIT_TYPE_POLICYLOAD and UTIL_AUDIT_TYPE_MAC_STATUS, and audit_log_user_avc_message for UTIL_AUDIT_TYPE_AVC. The result parameter is now passed to audit_log_user_message, set to 1 by default.

This change ensures proper argument order for ausearch utility as discussed in linux-audit/audit-userspace#394.

Fixes #387

### DIFF
--- a/src/util/audit.c
+++ b/src/util/audit.c
@@ -121,7 +121,11 @@ int util_audit_log(int type, const char *message, uid_t uid) {
         }
 
         if (audit_fd >= 0 && type != UTIL_AUDIT_TYPE_NOAUDIT) {
-                r = audit_log_user_avc_message(audit_fd, audit_type, message, NULL, NULL, NULL, uid);
+                if (type == UTIL_AUDIT_TYPE_AVC) {
+                        r = audit_log_user_avc_message(audit_fd, audit_type, message, NULL, NULL, NULL, uid);
+                } else {
+                        r = audit_log_user_message(audit_fd, audit_type, message, NULL, NULL, NULL, 1);
+                }
                 if (r <= 0)
                         return error_origin(-errno);
         } else {


src/util/audit.c
```diff
@@ -121,7 +121,11 @@ int util_audit_log(int type, const char *message, uid_t uid) {
         }
 
         if (audit_fd >= 0 && type != UTIL_AUDIT_TYPE_NOAUDIT) {
-                r = audit_log_user_avc_message(audit_fd, audit_type, message, NULL, NULL, NULL, uid);
+                if (type == UTIL_AUDIT_TYPE_AVC) {
+                        r = audit_log_user_avc_message(audit_fd, audit_type, message, NULL, NULL, NULL, uid);
+                } else {
+                        r = audit_log_user_message(audit_fd, audit_type, message, NULL, NULL, NULL, 1);
+                }
```
dvdhrm, line 128: Why are there two functions exposed by libaudit, anyway? Why doesn't `audit_log_user_avc_message()` just check whether the passed type is other than `AUDIT_USER_AVC` (or related AVC types), and then forward to `audit_log_user_message()`?

dvdhrm (approved):
Yeah, looks correct. Thanks a lot for the fix!

src/util/audit.c
Cropi: Perfectly valid question, I believe it's for historical reasons. The implementation of both functions is very similar but they provide a slightly different output. Documentation says:

- **audit_log_user_avc_message** This function should be used by all apps that are SELinux object managers.
- **audit_log_user_message** This function should be used by all console apps that do not manipulate accounts or groups.

dvdhrm: Btw., can you elaborate on the last argument to `audit_log_user_message`? The man-page is not very helpful. I guess this is about whether this is an audit-message about a successful / expected operation, or whether this is about a failed / unexpected operation?

I think passing `1` unconditionally might not be right here, though.

PR #389, opened 2025-04-16
util/sockopt: also check for ESRCH when the process has exited

bluca:
The kernel in 6.16 is changing the return value from EINVAL to ESRCH so that the particular case can be distinguished from generic errors:

https://lore.kernel.org/all/20250411-work-pidfs-enoent-v2-2-60b2d3bb545f@kernel.org/

### DIFF
--- a/src/util/sockopt.c
+++ b/src/util/sockopt.c
@@ -238,7 +238,7 @@ int sockopt_get_peerpidfd(int fd, int *pidfdp) {
                         return SOCKOPT_E_UNSUPPORTED;
                 if (errno == ENODATA)
                         return SOCKOPT_E_UNAVAILABLE;
-                if (errno == EINVAL)
+                if (errno == EINVAL || errno == ESRCH)
                         return SOCKOPT_E_REAPED;
 
                 return error_origin(-errno);


bluca:
Yeah it was reverted due to this code - the expected error is ESRCH when a process is gone, EINVAL doesn't make it distinguishable from other cases

dvdhrm:
> Yeah it was reverted due to this code - the expected error is ESRCH when a process is gone, EINVAL doesn't make it distinguishable from other cases

But isn't the expected error code `EINVAL`? That is what existing software uses, and what the kernel returns. I fully agree that `ESRCH` would have been a better choice back then.

If you change this upstream, I will inevitably have to deal with bug-reports of RHEL systems that update their kernel, but run LTS containers which suddenly fail.

Wouldn't it be easier to accept that `EINVAL` was used and just retain that? Or do you really insist on breaking backwards compatibility some time in the future?

Lastly, I would much more prefer if this entire condition would go away and I could just create pidfds for reaped tasks [1]. I don't know why this operation has to fail if the task is gone, given that the task can be reaped at any time after I created the pidfd as well.

[1] https://lore.kernel.org/lkml/20230807085203.819772-1-david@readahead.eu/

brauner:
> Lastly, I would much more prefer if this entire condition would go away and I could just create pidfds for reaped tasks [1]. I don't know why this operation has to fail if the task is gone, given that the task can be reaped at any time after I created the pidfd as well.

This is planned but not requires surgery in the socket specific code because we now also have `PIDFD_INFO_EXIT` which allows to retrieve the exit status from the pidfd (indefinitely) after the task has been reaped. The api is consistent in that you can only open a pidfd if you're guaranteed to get `PIDFD_INFO_EXIT` information. To make this work the socket code will need to ensure that an inode is allocated when stashing the peer struct pid so that it's guaranteed to have `PIDFD_INFO_EXIT` information available when handing out a pidfd for a reaped task. That's not difficult to do it just requires someone to do the work. Then we can allow this easily.

brauner:
> > Yeah it was reverted due to this code - the expected error is ESRCH when a process is gone, EINVAL doesn't make it distinguishable from other cases
> 
> But isn't the expected error code `EINVAL`? That is what existing software uses, and what the kernel returns. I fully agree that `ESRCH` would have been a better choice back then.
> 
> If you change this upstream, I will inevitably have to deal with bug-reports of RHEL systems that update their kernel, but run LTS containers which suddenly fail.

We've done that before and backported such changes to all LTS kernels for various apis. As far as we can tell you are the only project that specifically checks for this error code. If the pidfd api expands and takes additional flags the errors you see can change anyway and EINVAL might be completely overloaded just as it is in any other api.

brauner:
> > > Yeah it was reverted due to this code - the expected error is ESRCH when a process is gone, EINVAL doesn't make it distinguishable from other cases
> > 
> > 
> > But isn't the expected error code `EINVAL`? That is what existing software uses, and what the kernel returns. I fully agree that `ESRCH` would have been a better choice back then.
> > If you change this upstream, I will inevitably have to deal with bug-reports of RHEL systems that update their kernel, but run LTS containers which suddenly fail.
> 
> We've done that before and backported such changes to all LTS kernels for various apis. As far as we can tell you are the only project that specifically checks for this error code. If the pidfd api expands and takes additional flags the errors you see can change anyway and EINVAL might be completely overloaded just as it is in any other api.

In fact it was already before just not for the socket code. In general it could mean "reaped or not a thread-group leader".

bluca:
> > > Yeah it was reverted due to this code - the expected error is ESRCH when a process is gone, EINVAL doesn't make it distinguishable from other cases
> > 
> > 
> > But isn't the expected error code `EINVAL`? That is what existing software uses, and what the kernel returns. I fully agree that `ESRCH` would have been a better choice back then.
> > If you change this upstream, I will inevitably have to deal with bug-reports of RHEL systems that update their kernel, but run LTS containers which suddenly fail.
> 
> We've done that before and backported such changes to all LTS kernels for various apis. As far as we can tell you are the only project that specifically checks for this error code. If the pidfd api expands and takes additional flags the errors you see can change anyway and EINVAL might be completely overloaded just as it is in any other api.

Yeah in general we want to avoid EINVAL for very specific and expected cases, otherwise it's indistinguishable from user errors or programming errors, so it would be good to merge this change so that eventually we can change it back in the kernel too

dvdhrm:
> > If you change this upstream, I will inevitably have to deal with bug-reports of RHEL systems that update their kernel, but run LTS containers which suddenly fail.
> 
> We've done that before and backported such changes to all LTS kernels for various apis.

For real? You backport user-space adjustments to RHEL, SLES, Ubuntu, and whatever other distros big companies use for their deployments? You go through all that trouble knowing that at any time someone could just cause all your effort to be wasted because they tell you their internal systems suddenly failed with newer kernels, causing upstream changes to be reverted?

I personally do not want to spend my time on this, but I will not stand in your way. I can apply the patch, but I feel like this causes needless complications down the line when quality management needs to track down in some containers of a customer, etc.

> If the pidfd api expands and takes additional flags the errors you see can change anyway and EINVAL might be completely overloaded just as it is in any other api.

Yeah, but we only care for errors that can happen in our use-cases. Those will not change, just because you extend the upstream API. We know the situation we use pidfds in, and we rely on it to keep working in those conditions. Any other use-cases do not affect us, and you are free to change how things work there.

bluca:
> This is exactly what I mean: We do not call this on a non-thread-group-leader, so we do not care for the error-code overload.

How do you know that? It's a socket, there could be anything on the other side, no?

dvdhrm:
> > This is exactly what I mean: We do not call this on a non-thread-group-leader, so we do not care for the error-code overload.
> 
> How do you know that? It's a socket, there could be anything on the other side, no?

I think the SCM api provides the pidfd for the same pid as in the scm_pid field, which is always the thread-group leader of the peer. (not 100% sure, but I think that's how it works)

brauner:
> > > If you change this upstream, I will inevitably have to deal with bug-reports of RHEL systems that update their kernel, but run LTS containers which suddenly fail.
> > 
> > 
> > We've done that before and backported such changes to all LTS kernels for various apis.
> 
> For real? You backport user-space adjustments to RHEL, SLES, Ubuntu, and whatever other distros big companies use for their deployments? You go through all that trouble knowing that at any time someone could just cause all your effort to be wasted because they tell you their internal systems suddenly failed with newer kernels, causing upstream changes to be reverted?
> 
> I personally do not want to spend my time on this, but I will not stand in your way. I can apply the patch, but I feel like this causes needless complications down the line when quality management needs to track down in some containers of a customer, etc.

Yes, sure. We've also removed filesystems, system calls, deprecated general functionality. I mean it's very well documented that the rule is specifically about users complaining about api breaks not about not breaking api in general. There's tons of emails from Linus about this.

And most userspace projects we talk to which we know are affected are very forthcoming if we ask them to change what they're doing. Sometimes the api is new enough that there's almost no risk involved and sometimes we have to wait some time until we can pull the trigger but almost always we can eventually. Of course sometimes we change it and there's a breakage and then we revert.

dvdhrm:
> Yes, sure. We've also removed filesystems, system calls, deprecated general functionality.

Sure, but that's not really the same as removing actively used functionality. But lets just skip this.. so how about this:

I merge this, knowing that there are LTS distros without this change. So if you change the kernel, I rely on you forwarding this fix to all common LTS distros.

PR #392, opened 2025-04-30
build: add infrastructure for rust

dvdhrm:
Extend the Meson build system to compile an internal Rust library and link that into `libbus.a`, our internal utility library that is linked into all dbus-broker binaries.

The individual commits have extensive information on, and background for, the individual changes. To summarize, we now use the builtin Rust support of Meson to compile a static library that uses no external dependencies but standard Rust `core` and `alloc`, as well as the runtime-integration of `std`. This avoids the previous dance via `meson-cargo`, and it makes the build fully independent of `Cargo`.

This series merely extends the build system, but adds no meaningful Rust code. This will be added in other PRs.

### DIFF
--- a/.github/workflows/meson-tests.yml
+++ b/.github/workflows/meson-tests.yml
@@ -36,31 +36,29 @@ jobs:
         # since it clashes with sd-bus used by the launcher. This test also
         # builds documentation and related resources.
         - id: "release"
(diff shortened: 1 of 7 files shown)


dvdhrm:
(replaced by #399)

Issue #390, opened 2025-04-25
Implement standard Debug Stats interface data


cooperlees:
Hello,

Could we possibly get the (recently) added DBus Stats interface data implemented for dbus-broker?
- https://dbus.freedesktop.org/doc/dbus-specification.html#message-bus-debug-stats-interface
  - Added in Revision 0.43 - 2024-10-29

I found this was unimplemented by dbus-broker as I wanted to collect per UID account/resource metrics via a daemon we run to monitor systemd via Dbus we call [monitord](https://github.com/cooperlees/monitord). This is written in rust and uses the [zbus](https://github.com/dbus2/zbus) crate and when I looked at zbus's Stats implementation it didn't work. We then noticed "fixing it" the data returns from dbus-broker seems specific to dbus-broker itself and not the standard.

dvdhrm:
I think most of this information we have available and could return. I am curious, though, what do you do with this data?

Also note that this is a debug interface, so I wouldn't use it for live monitoring. At least in the case of dbus-broker, it is quite a heavy call that stalls the bus for a short time (<1ms though).

cooperlees:
> what do you do with this data?

We are hitting UID quotas and want to watch the usage as we tune things to get an idea who / what is talking to dbus the most. I would look to agregate it from our hosts to get an idea across our different server types the DBus usage. It's a big unknown at the moment and a new hardware type is constantly getting resource DBus errors like:

```
Peer :1.92799 is being disconnected as it does not have the resources to perform an operation.
```

It has thousands and thousands of systemd units, so we are expecting the systemd DBus call returns to be large due to that.

> stalls the bus for a short time (<1ms though).

Interesting to know. Thanks for sharing. I would be looking to poll periodically, I would have to reduce it due to cost then.

ryantimwilson:
I took a pass at adding some of the "basic" standard metrics which is basically just summing up dbus-broker's per-namespace metrics: https://github.com/bus1/dbus-broker/pull/393.

The per-namespace metrics are probably most useful if we want to find the service that's communicating most with dbus, at the risk that the metric format is dbus-broker implementation dependent.

I didn't add the Peak* metrics as it seems those require adding additional tracking metrics and those don't seem as useful (but happy to be wrong).

Let me know if I'm way off here.

cooperlees:
> The per-namespace metrics are probably most useful if we want to find the service that's communicating most with dbus, at the risk that the metric format is dbus-broker implementation dependent.

[...]

PR #393, opened 2025-05-04
Add some standard stats to Debug.GetStats method

ryantimwilson:
This adds standard fields as expected by the org.freedesktop.DBus.Debug.Stats.GetStats method:
- Serial
- ActiveConnections
- IncompleteConnections
- MatchRules
- BusNames

Since these fields are essentially summing the per-namespace stats over all peers, these stats are easy to gather. However, this commit does not expose Peak* stats as these would require tracking max matches and bus names over the lifetime of dbus-broker. Arguably, the # of connections, match rules and bus names are more important.

See #390

### DIFF
--- a/src/bus/driver.c
+++ b/src/bus/driver.c
@@ -2127,7 +2127,7 @@ static int driver_method_get_all(Peer *peer, const char *path, CDVar *in_v, uint
         return 0;
 }
 
-static void driver_append_peer_accounting_stats(CDVar *v, Peer *peer) {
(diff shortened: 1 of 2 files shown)


dvdhrm (approved):
This looks good! Some minor style issues, but I will amend this myself.

I wonder, though, whether we should just track these counters at all times, and then add the `Peak*` values? I think there is little downside, and the we would avoid deviations from the spec.

Regardless, I will merge this and then see whether I can adjust it in a followup. Thanks a lot!

dvdhrm:
I decided to just rebase your commit on top of my changes to the resource counters. See #403 for details! I will close this PR, lets move discussion there.

Note that I amended you commit, so please let me know if you disagree with the change. And thanks again for starting this!

PR #394, opened 2025-05-05
Silence new GCC warnings

dvdhrm:
Again some warnings of GCC that will break builds with `-Werror` but are false positives. See each one for details.


dvdhrm:
(code fuzzers have been disabled and fail for unrelated reasons)

evverx:
> code fuzzers have been disabled and fail for unrelated reasons

The failures should be addressed in https://github.com/google/oss-fuzz/pull/13286

dvdhrm:
Thanks a lot for working on this!

PR #397, opened 2025-05-09
ci: rebase on archlinux

dvdhrm:
Switch the base image again from Fedora to Arch Linux. This requires us to disable SELINUX in the unit-tests, but those will be run by the packit integration, so we should be fine.

The advantages are:

 * Image builds take *significantly* less time: 10min for Fedora, 3min for Arch Linux (on the same infrastructure).

 * Images are significantly smaller in size: 6GiB for Fedora, 3GiB for ArchLinux (with the same packages).

We also get rolling releases, allowing us to control at which point we can update. We do not expose the CI system to untrusted inputs, anyway (nor do we pass sensitive data to it).

### DIFF
--- a/.github/workflows/meson-tests.yml
+++ b/.github/workflows/meson-tests.yml
@@ -36,31 +36,29 @@ jobs:
         # since it clashes with sd-bus used by the launcher. This test also
         # builds documentation and related resources.
         - id: "release"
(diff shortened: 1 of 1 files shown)


dvdhrm:
Note that we still run some tests on Fedora, some on Ubuntu, etc. This just switches the default container to one based on ArchLinux. We use this container for any non-standard builds like `-DNDEBUG`, `-O0`, `fallback`, `32bit`, etc.

PR #399, opened 2025-05-09
build: add infrastructure for rust (v2)

dvdhrm:
Extend the Meson build system to compile an internal Rust library and link that into libbus.a, our internal utility library that is linked into all dbus-broker binaries.

The individual commits have extensive information on, and background for, the individual changes. To summarize, we now use the builtin Rust support of Meson to compile a static library that uses no external dependencies but standard Rust core and alloc, as well as the runtime-integration of std. This avoids the previous dance via meson-cargo, and it makes the build fully independent of Cargo.

This series merely extends the build system, but adds no meaningful Rust code. This will be added in other PRs.

(v2: rebased on the recent CI changes)

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -48,6 +48,9 @@ sources_bus = [
         'util/user.c',
 ]
 
+bindgen_bus = [
+]
+
 deps_bus = [
         dep_cdvar,
         dep_clist,
@@ -110,6 +113,40 @@ else
         ]
 endif
 
+bindgen_args = [
+        '--formatter=prettyplease',
+        '--rust-edition=' + rust_edition,
+]
+
+bindgen_generated = []
+
+foreach i : bindgen_bus
+        bindgen_generated += mod_rust.bindgen(
+                args: bindgen_args + [
+                        '--allowlist-file=.*/' + i,
+                ],
+                bindgen_version: bindgen_version,
+                dependencies: deps_bus,
+                include_directories: incs_bus,
+                input: i,
+                # `foo-bar.h` -> `foo_bar.rs`
+                output: i.substring(0, -2).underscorify() + '.rs',
+        )
+endforeach
+
+rust_bus = static_library(
+        'rbus',
+        structured_sources(
+                [
+                        'lib.rs',
+                ],
+                {
+                        'generated': bindgen_generated,
+                },
+        ),
+        rust_abi: 'c',
+)
+
 static_bus = static_library(
         'bus-static',
         sources_bus,
@@ -118,6 +155,7 @@ static_bus = static_library(
                 '-fno-common',
         ],
         dependencies: deps_bus,
+        link_with: rust_bus,
         pic: true,
 )
 
@@ -175,6 +213,12 @@ endif
 #
 # target: test-*
 #
+
+mod_rust.test(
+        'rbus-tests',
+        rust_bus,
+        suite: 'unit',
+)
 
 test_kwargs = {
         'dependencies': dep_bus,
--- a/src/meson.build
+++ b/src/meson.build
@@ -114,8 +114,10 @@ else
 endif
 
 bindgen_args = [
-        '--formatter=prettyplease',
-        '--rust-edition=' + rust_edition,
+        '--formatter', 'prettyplease',
+        '--rust-edition', rust_edition,
+        '--rust-target', rust_msrv,
+        '--use-core',
 ]
 
 bindgen_generated = []
--- a/src/meson.build
+++ b/src/meson.build
@@ -115,10 +115,15 @@ endif
 
 bindgen_args = [
         '--formatter', 'prettyplease',
-        '--rust-edition', rust_edition,
-        '--rust-target', rust_msrv,
         '--use-core',
 ]
+
+if bindgen_version.version_compare('>=0.71')
+        bindgen_args += [
+                '--rust-edition', rust_edition,
+                '--rust-target', rust_msv,
+        ]
+endif
 
 bindgen_generated = []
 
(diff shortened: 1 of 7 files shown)


evverx:
Looking at https://github.com/bus1/dbus-broker/actions/runs/14931302421 it fails to build with the OSS-Fuzz toolchain as expected. It should be possible to switch to the rust builder image and install bindgen-cli there to get it to work for the time being but generally projects written in both C and Rust built with meson aren't natively supported there. I wonder if there are plans to keep fuzzing the C code and start fuzzing the fuzzable Rust code once it's introduced or should it be disabled on OSS-Fuzz until it's all settled?

bonzini:
The change in [bc02e79](https://github.com/bus1/dbus-broker/pull/399/commits/bc02e79d296160678f57c4059aeedf3faee07afc) should not be needed in 1.8.1, hopefully, so maybe just change the version comparison once the release is out.

src/meson.build
```diff
@@ -228,6 +228,14 @@ mod_rust.test(
         suite: 'unit',
 )
 
+if meson.version().version_compare('>=1.8')
+        mod_rust.doctest(
+                'rbus-doctests',
+                rust_bus,
+                suite: 'unit',
```
bonzini, line 239: I suggest using a separate suite here. I'm not sure how your CI works, but some projects do not copy over the individual .o files from the `build` stage to the `test` stage. That may prevent you from running the doctests in the `test` stage, and having a separate suite helps moving them to the `build` stage.

dvdhrm: Hmmm, `rustdoc` does not allow to split the build stage from the test stage, does it? So you have to call `rustdoc --test [...]` at the time of running the tests, which then needs access to the crate it documents, as this is usually linked into the doc-tests. I see.

I am kinda leaning towards just making doctests build-time optional, then it is easy to disable when anyone runs into such problems.

But putting it into a separate suite is definitely nice, so it can be easily deselected, given its runtime requirements. Thanks for the hint!

bonzini: > Hmmm, rustdoc does not allow to split the build stage from the test stage, does it? 

Yes that's the issue. I meant build and test stage as CI concepts (I'm more familiar with Gitlab but GitHub should have something similar).

> I am kinda leaning towards just making doctests build-time optional, then it is easy to disable when anyone runs into such problems.

Yes that also makes sense and would also be a workaround for any issues with cross toolchain files.

dvdhrm: Sounds good, thanks a lot!

evverx:
I think it would be great to add things like clippy and rustfmt eventually. Looking at https://github.com/mesonbuild/meson/pull/13914 it seems `ninja clippy` is implemented so it can probably be added to the actions right away. `rustfmt` is in the works as far as I can see (https://github.com/mesonbuild/meson/pull/13932). FWIW it complains
```diff
Diff in /dbus-broker/src/rlib.rs at line 20:
 #[allow(dead_code)]
 #[allow(non_camel_case_types)]
 #[allow(non_snake_case)]
-pub mod generated {
-}
+pub mod generated {}
 
 #[cfg(test)]
 mod test {
Diff in /dbus-broker/src/rlib.rs at line 28:
     // Simple dummy to show the test-suite in the test results, even if
     // other tests are conditionally disabled.
     #[test]
-    fn availability() {
-    }
+    fn availability() {}
 }
 

```

bonzini:
You can also check QEMU's master branch for how it enables "meson devenv cargo FOO" (where FOO is usually fmt). We use it to run rustfmt in CI.

evverx:
Thanks! I just took a quick look but it seems it should probably be possible to borrow something from their CI: https://github.com/qemu/qemu/blob/bc98ffdc7577e55ab8373c579c28fe24d600c40f/.gitlab-ci.d/static_checks.yml#L54. I saw MIRI in the docs as well.

dvdhrm:
> I'm going to fuzz dbus-broker downstream (by analogy with dbus-daemon [google/oss-fuzz#8699](https://github.com/google/oss-fuzz/pull/8699)) so I'll go ahead and close [google/oss-fuzz#13299](https://github.com/google/oss-fuzz/pull/13299). When/if it breaks there it should be possible to disable it by removing the project.

Thanks a lot! I merged this PR so we have a baseline Rust integration that we can base other work like oss-fuzz on. If your branch works with this PR, I'd be glad to see it merged in oss-fuzz. I know that fuzzing Rust code will take more effort, but we should at least try to keep fuzzing the message-generator as we did before.

dvdhrm:
> > I don't intend to enable rustfmt
> 
> Various IDEs run `rustfmt` automatically when they see Rust code so I think it would be better to add `.rustfmt.toml` with `ignore` (or something like that) anyway to avoid changes where files get reformatted in unrelated places accidentally.

An IDE that reformats an entire file when you touch it? Sounds horrid :-/

I think it would be easier to just put `#![rustfmt::skip]` into rlib.rs. Does that sounds ok?

> It's also possible to skip code granularly (which I think would be a better option in terms of utilizing rustfmt in places where it makes sense). There are also options that can be set to control `brace_style` for example but they are in nightly only.

I have tried that before. For instance, in r-efi (https://crates.io/crates/r-efi) we used to do this, but we just kept adding annotations and it become really ugly.

Note that we are not alone on this. The official rust style-guide explicitly leaves decision on what a "short item" is up to an implementer (https://doc.rust-lang.org/nightly/style-guide/). And there are many, many discussion in the rustfmt tracker about this.

We explicitly follow the Rust style guide (I think?), I just dislike the decisions rustfmt did when implementing it (which are in no way mandated by the style guide).

> > If your branch works with this PR, I'd be glad to see it merged in oss-fuzz
> 
> It works in the sense that dbus-broker compiles. `RUSTFLAGS` aren't set up properly and things like that but it should be fine for now I guess.

I see. Thanks for tracking it!

PR #406, opened 2025-06-11
ci: run coverity analysis regularly

dvdhrm:
This series provides coverity-helpers in `./Makefile` which allows to locally build and run the coverity tools. Additionally, a workflow is provided to regularly (and on demand) run coverity on `main`.

@mrc0mmand, @evverx, let me know what you think.

### DIFF
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -1,0 +1,48 @@
+#
+# Coverity Analysis
+#
+# This workflow analyzes the project via the Coverity Scan Analysis Tool, and
+# uploads the final artifacts to the Coverity servers for further analysis.
+#
+# This workflow can be triggered manually. It is also run in regular intervals.
+# Note that Coverity limits the amount of builds allowed, so ensure you do not
+# exceed this with manual triggers.
+#
+
+name: "Coverity Analysis"
+
+on:
+  schedule:
+  - cron:  '0 0 * * *' # daily at midnight
+  workflow_dispatch:
+
+defaults:
+  run:
+    shell: "bash"
+
+permissions:
+  contents: read
+
+jobs:
+  analysis:
+    name: "Coverity Analysis"
+
+    if: github.repository == 'bus1/dbus-broker'
+
+    container:
+      image: "ghcr.io/readaheadeu/rae-ci-ubuntu:latest"
+      options: --user root
+    env:
+      COVERITY_EMAIL: "no-reply@readahead.eu"
+      COVERITY_TOKEN: ${{ secrets.DEPLOY_COVERITY_TOKEN }}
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -59,6 +62,98 @@ $(BUILDDIR)/%/:
...
+.PHONY: coverity-scan
+coverity-scan: $(COVERITY_P_SCAN)
+	@echo -e "\033[33;1mcoverity: run full analysis\033[0m"
+	@rm -rf "$(COVERITY_P_BUILD)" "$(COVERITY_P_INT)"
+	@mkdir -p "$(COVERITY_P_BUILD)" "$(COVERITY_P_INT)"
+	@meson \
+		setup \
+		--buildtype "debugoptimized" \
+		--warnlevel "2" \
+		-D "launcher=true" \
```
evverx, line 130: I think it would be great to flip `audit`, `selinux` and the other options (apart from the tests probably) to cover as much code as possible.

evverx (commented):
Thanks! @mrc0mmand is a coverity expert so I'll leave it to @mrc0mmand to review this.

FWIW I closed that coverity issue because as far as I understand the rust infrastructure is on the way so I wanted to avoid adding one more build that would need fixing. I'm moving the fuzzing stuff downstream because of that as well. Codebases written in C and Rust aren't well-supported there (in the sense that coverage reports and fuzz introspector reports are messed up, it's necessary to manually do things `cargo fuzz` does, projects like suricata basically maintain the rust toolchain and thing like that).

Makefile
dvdhrm: This is currently built on ubuntu. Does this ship with selinux libraries?

evverx: As far as I can see Ubuntu Noble ships `libselinux1-dev` and `libapparmor-dev` so it should be possible to flip those features. Let me double check.

.github/workflows/coverity.yml
```diff
@@ -0,0 +1,49 @@
...
+  contents: read
+
+jobs:
+  analysis:
+    name: "Coverity Analysis"
+
+    if: github.repository == 'bus1/dbus-broker'
+
+    container:
+      image: "ghcr.io/bus1/dbrk-ci-ubuntu:latest"
```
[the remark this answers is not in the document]
dvdhrm: You can actually just resolve it as URL: https://ghcr.io/bus1/dbrk-ci-ubuntu

It is defined in the `bus1/cabuild` repository, where we build all the utility images.

Makefile
dvdhrm: FYI: I adjusted almost all CI runs to use the ubuntu image and build with all options enabled. Thanks for the hint!

PR #407, opened 2025-06-18
ci: major rework

dvdhrm:
This is a major rework of the continuous integration. The aim is to
centralize any code build & analysis hooks in a single workflow, and
call other workflows from it. This allow serializing them, if desired.

In particular, this re-introduces the `ci.yml` workflow, which is meant
to be the main workflow run on pushes or PRs. It first runs a set of
basic tests, which should catch most simple mistakes found in PRs, like
typos or simple build failures. Only if those succeed, it will spawn the
extended test-suite.

This setup conserves CI resources at minimal cost. The basic tests run
really fast, and thus the experience for most PRs stays the same. But we
avoid running the test matrix for simple mistakes like missing commit
hunks, typos, failed force-pushes, etc. Given that the CI resources are
limited for the `bus1` organization, this should overall increase CI
performance, because machines become available earlier.

As a second step, this PR extends the CI considerably with the following
new features:

- By default, we now use CLANG with LLD for better performance and
  code analysis.
- We make use of native meson support for `debug`, `werror`, and
  `errorlogs`, ensuring those values are properly passed along.
- We run `clippy` on the Rust code-base for static analysis.
- We use `ubuntu` based containers for most runs to ensure we can build
  with both SELinux *and* AppArmor. No other distro allows this.
- We add runs which use native ARM64 machines to test on non-x86
  hardware. This requires container images for arm64, which was
  implemented separately.
- We invoke the CodeQL workflow from within this workflow, to ensure it
  is also guarded by the basic checks.
- We now run some checks with LTO and thinLTO.
- Instead of the `-m32` hack, we now use proper cross-compilation via
  meson cross files for native 32-bit tests.
- We now run UBSAN and ASAN in CI.
- Valgrind integration is reworked and now run as proper individual
  test.


dvdhrm:
I would love to be able to guard packit by the same logic: Invoke it only if the basic tests succeed. But packit is currently triggered by github-hooks, and I haven't figured out how we can change this to use our own manual triggers.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,372 @@
...
+
+        # Run with address-sanitizers enabled.
+        - id: "asan"
+          name: "Ubuntu-x86_64-clang-release"
+
+          cc: "clang"
+          cc_ld: "lld"
+          image: "ghcr.io/readaheadeu/rae-ci-ubuntu:latest"
+          setupargs: >-
+            -D b_sanitize=address
```
evverx, line 288: `-D b_sanitize=address` affects the C part only. https://doc.rust-lang.org/beta/unstable-book/compiler-flags/sanitizer.html#addresssanitizer would have to be passed via `RUSTFLAGS`. It's in nightly only though.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,372 @@
...
+      run: meson test -C "./build"
+
+  #
+  # A release build running through valgrind. We disable all features
+  # that currently do not support valgrind runs (in particular sd-bus in
+  # the launcher).
+  #
+  ext_valgrind:
+    name: "Ext: valgrind @ Arch-x86_64-clang-release"
+    needs: ext_guard
```
evverx, line 337: On my machine the testsuite fails under Valgrind. I'm not sure what's different though
```sh
48/51 dbus-broker:dbus-broker / FD Stream Constraints FAIL             4.59s   killed by signal 6 SIGABRT
...
==3728== Syscall param sendmsg(mmsg[4].msg_hdr.msg_control) points to uninitialised byte(s)                                                                   
==3728==    at 0x49538A3: sendmmsg (sendmmsg.c:30)                                                                                                            
==3728==    by 0x42912B: socket_dispatch_write (socket.c:848)                                                                                                 
==3728==    by 0x429456: socket_dispatch (socket.c:953)                                                                                                       
==3728==    by 0x42366B: connection_dispatch (connection.c:225)                                                                                               
==3728==    by 0x41A900: peer_dispatch_connection (peer.c:37)                                                                                                 
==3728==    by 0x41B5DB: peer_dispatch (peer.c:201)                                                                                                           
==3728==    by 0x42A3C5: dispatch_context_dispatch (dispatch.c:343)                                                                                           
==3728==    by 0x4022C1: broker_run (broker.c:203)                                                                                                            
==3728==    by 0x401260: run (main.c:280)                                                                                                                     
==3728==    by 0x4013FF: main (main.c:311)                                                                                                                    
==3728==  Address 0x4a57a5c is 28 bytes inside a block of size 32 alloc'd                                                                                     
==3728==    at 0x483EB26: malloc (vg_replace_malloc.c:446)                                                                                                    
==3728==    by 0x42A5A5: fdlist_new_with_fds (fdlist.c:32)                                                                                                    
==3728==    by 0x42A67A: fdlist_new_consume_fds (fdlist.c:60)                                                                                                 
==3728==    by 0x428C6E: socket_recvmsg (socket.c:688)                                                                                                        
==3728==    by 0x428E1F: socket_dispatch_read (socket.c:731)                                                                                                  
==3728==    by 0x429445: socket_dispatch (socket.c:950)                                                                                                       
==3728==    by 0x42366B: connection_dispatch (connection.c:225)                                                                                               
==3728==    by 0x41A900: peer_dispatch_connection (peer.c:37)                                                                                                 
==3728==    by 0x41B5DB: peer_dispatch (peer.c:201)                                                                                                           
==3728==    by 0x42A3C5: dispatch_context_dispatch (dispatch.c:343)                                                                                           
==3728==    by 0x4022C1: broker_run (broker.c:203)                                                                                                            
==3728==    by 0x401260: run (main.c:280)                                                                                                                     
==3728==  Uninitialised value was created by a heap allocation                                                                                                
==3728==    at 0x483EB26: malloc (vg_replace_malloc.c:446)                                                                                                    
==3728==    by 0x42A5A5: fdlist_new_with_fds (fdlist.c:32)                                                                                                    
==3728==    by 0x42A67A: fdlist_new_consume_fds (fdlist.c:60)                                                                                                 
==3728==    by 0x428C6E: socket_recvmsg (socket.c:688)                                                                                                        
==3728==    by 0x428E1F: socket_dispatch_read (socket.c:731)                                                                                                  
==3728==    by 0x429445: socket_dispatch (socket.c:950)                                                                                                       
==3728==    by 0x42366B: connection_dispatch (connection.c:225)                                                                                               
==3728==    by 0x41A900: peer_dispatch_connection (peer.c:37)                                                                                                 
==3728==    by 0x41B5DB: peer_dispatch (peer.c:201)                                                                                                           
==3728==    by 0x42A3C5: dispatch_context_dispatch (dispatch.c:343)                                                                                           
==3728==    by 0x4022C1: broker_run (broker.c:203)                                                                                                            
==3728==    by 0x401260: run (main.c:280)                                                                                                                     
==3728==                                             
```
`Driver API` fails with the same backtrace.

dvdhrm: Is this new? We always supported valgrind runs, it was just not as prominent as it is now.

Can you maybe move this into a separate report? It looks like missing annotations, but I haven't fully traced it, yet.

dvdhrm: I finally found the time to look into this. It seems that we never clear trailing cmsg padding when sending an odd-number of FDs on 64bit systems. (or even number? depends on the size of the cmsg-header, I guess).

I don't think this is a problem, since the kernel copies those FDs manually, anway. However, I fixed this now. See #414.

Thanks a lot for the report!

evverx: I forgot about this. Sorry!

> I don't think this is a problem, since the kernel copies those FDs manually, anway. However, I fixed this now

Thanks! I think Valgrind should keep producing other false positives caused by pidfds (with `--track-fds`)
```sh
==291== File descriptor 9: /proc/sys/kernel/ngroups_max is already closed
...
```
but that's not something that should be addressed upstream. Those things can be suppressed downstream (and should be fixed in Valgrind upstream ideally).

PR #426, opened 2026-02-03
apparmor: synchronize with dbus-daemon(1)

dvdhrm:
This synchronizes the apparmor module with dbus-daemon(1). In particular, the send checks are completely reworked to just do literal checks on the addressed recipient, rather than using a full name-walk.

See the individual commit messages for details.


dvdhrm:
(packit builds fail for unrelated reasons)

dvdhrm:
(update: rebased on the packit fix; added commit to include destination/source in the audit logs)

PR #437, opened 2026-06-11
sockopt: handle EMFILE/ENFILE in sockopt_get_peerpidfd gracefully

### LINKED ISSUE #435: dbus-broker-launch shuts down when hitting user soft fd limit
Any process spamming connections to the session bus will cause immediate crash of dbus-broker, and logout of the user session.

Note that this means you can crash the desktop from within a flatpak sandbox, as the flatpak dbus proxy simply forwards connections immediately to the host session bus.

Simple reproducer:
https://codeberg.org/jsimmons/crash-dbus

```
May 14 21:31:15 zenzen dbus-broker-launch[2844]: ERROR sockopt_get_peerpidfd @ ../dbus-broker-37/src/util/sockopt.c +244: Too many open files
May 14 21:31:15 zenzen dbus-broker-launch[2844]:       peer_new_with_fd @ ../dbus-broker-37/src/bus/peer.c +290
May 14 21:31:15 zenzen dbus-broker-launch[2844]:       listener_dispatch @ ../dbus-broker-37/src/bus/listener.c +54
May 14 21:31:15 zenzen dbus-broker-launch[2844]:       dispatch_context_dispatch @ ../dbus-broker-37/src/util/dispatch.c +344
May 14 21:31:15 zenzen dbus-broker-launch[2844]:       broker_run @ ../dbus-broker-37/src/broker/broker.c +209
May 14 21:31:15 zenzen systemd[2826]: Got disconnect on API bus.
```

It seems like the intent for this code is that going beyond some quota leads to new connections being dropped, however dbus-broker-launch doesn't increase its soft fd limit from 1024 so opening a large number of connections will instantly kill the session bus.

https://github.com/bus1/dbus-broker/blob/404f092d3742c57b58fba5169e64c2999f00f794/src/bus/listener.c#L46-L52

tianming-1996:
## Summary

- Fix a crash in `dbus-broker` when a client's `SO_PEERPIDFD` query fails due to
  file-descriptor exhaustion (`EMFILE`/`ENFILE`).
- Treat fd-exhaustion errors as `SOCKOPT_E_UNAVAILABLE`, matching the semantic
  that a missing PID fd is optional metadata — the caller already handles this
  gracefully (drops the new connection), so the broker no longer aborts.
- Add a regression test that deliberately exhausts the fd table in a subprocess and
  verifies the correct return value.

## Root Cause

When a connection spam fills the session bus, `getsockopt(SO_PEERPIDFD)` returns
`EMFILE` ("too many open files") because no spare fd slot exists for the kernel
to return the new pidfd. Previously this fell through to `error_origin(-errno)`,
propagating a fatal error that crashed the broker.

## Changes

| File | Change |
|------|--------|
| `src/util/sockopt.c` | Handle `EMFILE`/`ENFILE` in `sockopt_get_peerpidfd()` → return `SOCKOPT_E_UNAVAILABLE` |
| `src/util/test-sockopt.c` | Add `test_peerpidfd_fd_exhaustion` regression test |

## Test Plan

- Unit test `test_peerpidfd_fd_exhaustion` verifies `sockopt_get_peerpidfd()`
  returns `SOCKOPT_E_UNAVAILABLE` when fd table is full.
- Full CI test suite (`meson test`) validates no regressions.

## Build Notes

Local full-build requires: `rust >= 1.63`, `jq >= 1.6`, `libexpat-dev`,
`libaudit-dev`. The core logic change (2 lines) is C11-compatible and
syntax-checked with `gcc -std=c11`.

Fixes #435

### DIFF
--- a/src/util/sockopt.c
+++ b/src/util/sockopt.c
@@ -240,6 +240,8 @@ int sockopt_get_peerpidfd(int fd, int *pidfdp) {
                         return SOCKOPT_E_UNAVAILABLE;
                 if (errno == EINVAL || errno == ESRCH)
                         return SOCKOPT_E_REAPED;
+                if (errno == EMFILE || errno == ENFILE)
+                        return SOCKOPT_E_UNAVAILABLE;
 
                 return error_origin(-errno);
         }
--- a/src/util/test-sockopt.c
+++ b/src/util/test-sockopt.c
@@ -4,7 +4,9 @@
 
 #undef NDEBUG
 #include <c-stdaux.h>
+#include <fcntl.h>
 #include <stdlib.h>
+#include <sys/resource.h>
 #include <sys/socket.h>
 #include <sys/types.h>
 #include <sys/un.h>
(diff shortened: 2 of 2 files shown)


dvdhrm (changes requested):
Thanks for the PR!

Our aim is to make SO_PEERPIDFD mandatory, once the MSV of the kernel is raised sufficiently. So this fix does not really solve the issue, it just makes it work like it worked before.

I am going to close this PR and move discussion to #435, so we can figure out a way to fix this without disabling the peer-pidfds.

PR #438, opened 2026-06-19
apparmor: Implement audit log fallback

aleasto:
If we're unable to send kernel audits then print the audit message to the log instead. Matches the behaviour of dbus-daemon.

### DIFF
--- a/src/util/apparmor.c
+++ b/src/util/apparmor.c
@@ -15,11 +15,13 @@
 #include "util/apparmor.h"
 #include "util/audit.h"
 #include "util/error.h"
+#include "util/log.h"
 #include "util/ref.h"
 #include "util/string.h"
 
 struct BusAppArmorRegistry {
         _Atomic unsigned long n_refs;
+        Log *log;
         char *bustype;
         char fallback_context[];
 };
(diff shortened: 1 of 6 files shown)


src/util/apparmor.c
```diff
@@ -179,9 +193,16 @@ static int bus_apparmor_log(
...
-        if (r != UTIL_AUDIT_E_UNAVAILABLE) // XXX: use a log fallback
+        if (r != UTIL_AUDIT_E_UNAVAILABLE)
                 return error_fold(r);
 
+        if (bus_apparmor_log_context) {
+                log_append_here(bus_apparmor_log_context, LOG_NOTICE, 0, NULL);
+                r = log_commitf(bus_apparmor_log_context, "%s", message);
+                if (r)
+                        return error_fold(r);
+        }
```
dvdhrm, line 194: Can we make `bus_apparmor_log_context` a member of `BusAppArmorRegistry`? I don't think we need the global state, do we? That is, make `bus_apparmor_registry_new()` take the log context, rather than adding the global init/deinit.

SELinux does not pass the context to the callbacks, so we need the global init/deinit. But I would love to avoid it for apparmor, unless necessary!

dvdhrm (changes requested):
Great! Thanks a lot! One comment inline, but conceptually I like it!

src/util/apparmor.c
aleasto: Done.

dvdhrm (approved):

dvdhrm:
Thanks a lot! This looks great! Applied.

PR #442, opened 2026-06-28
util/sockopt: fix peer group deduplication

SABITHSAHEB:
While reading through `sockopt_get_peergroups()` I noticed the peer group
list is never actually deduplicated. After `SO_PEERGROUPS`, the list is
sorted and the dedup loop compacts the unique entries to the front, but
`n_gids` is never updated, so the original (pre-dedup) count is returned.
A duplicate GID is left at the tail and the count is too large.

It's easy to hit, since the primary GID at `gids[0]` is often also one of
the supplementary groups, e.g. `[3, 5, 7]` with primary GID `5`:

### DIFF
--- a/src/util/sockopt.c
+++ b/src/util/sockopt.c
@@ -121,6 +121,7 @@ int sockopt_get_peergroups(int fd, Log *log, uid_t uid, gid_t primary_gid, gid_t
                                 if (gids[i] != gids[j])
                                         gids[++j] = gids[i];
                         }
+                        n_gids = j + 1;
 
                         if (gidsp) {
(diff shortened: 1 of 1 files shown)


dvdhrm (approved):
Nice catch!

We properly sort and perform the deduplication. Hence, the trailing data can only contain duplicates, right? So the original code returned correct data, but might have reported trailing duplicates. I don't think this needs any special treatment.

Thanks a lot for the patch!

(I will drop the `.wraplock` update when applying; not sure why it was included)

dvdhrm:
(merged manually)

PR #450, opened 2026-07-01
dbus/message: avoid alloca() when parsing the body signature

SABITHSAHEB:
The body-signature parser allocated its CDVarType scratch array with
alloca(), sized from the message signature. alloca() has no failure mode,
is a stack-clash primitive, and is routinely flagged by the static
analyzers run in CI (CodeQL, Coverity) -- undesirable on a path that
processes untrusted message input.

A D-Bus signature is limited to 255 characters and each character maps to
exactly one CDVarType entry, so a fixed 256-entry (1 KiB) on-stack array
is always large enough. Use such an array and reject up front any
signature that would not fit. The previous bound was only checked via
c_assert(), which is compiled out in NDEBUG builds; the explicit check
keeps it enforced in all builds while dropping the variable-length
allocation entirely. No behavior changes for valid messages.

### DIFF
--- a/src/dbus/message.c
+++ b/src/dbus/message.c
@@ -368,18 +368,23 @@ static int message_parse_header(Message *message, MessageMetadata *metadata) {
 static int message_parse_body(Message *message, MessageMetadata *metadata) {
         _c_cleanup_(c_dvar_deinit) CDVar v = C_DVAR_INIT;
         const char *signature = metadata->fields.signature;
+        CDVarType types[256];
         size_t i, n_signature, n_types;
-        CDVarType *t, *types;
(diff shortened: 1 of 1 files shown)


dvdhrm:
Thanks for the patch! You describe quite clearly that alloca() works fine in this code-path, so I am unsure why you want to change it?

SABITHSAHEB:
@dvdhrm Thanks for the review and for pointing out that alloca() works fine here. I understand your perspective, and I appreciate you looking into this so promptly. Let me know if you'd like me to make any further changes.

dvdhrm:

PR #452, opened 2026-07-03
util/apparmor: prevent integer overflow in message query buffer alloc…

SABITHSAHEB:
This patch makes the AppArmor query builder safer by checking the total buffer size before allocating memory. In practice, that closes an integer-overflow path that could otherwise lead to an undersized allocation and unsafe string copies afterward. It’s a small, focused hardening change that improves reliability without changing the feature’s behavior.


dvdhrm:

PR #453, opened 2026-07-03
util/string: reject negative values in unsigned integer parsers

SABITHSAHEB:
Improve util_strtou32 and util_strtou64 to validate negative signs.
Only the "-1" sentinel is allowed (mapping to UINT32_MAX/UINT64_MAX
for unlimited capacity), while all other negative numbers (such as
"-2" or "-100") are rejected. 

This prevents integer overflow bypasses when parsing configurations
or command-line limits.


dvdhrm:

PR #454, opened 2026-07-04
util/string: reject negative inputs in unsigned integer parsers

SABITHSAHEB:
Reject negative inputs in util_strtou32 and util_strtou64. Currently, passing negative values like "-1" results in a silent wrap-around to ULONG_MAX because of how strtoul/strtoull operate. 

This patch hardens both unsigned helpers to explicitly return UTIL_STRING_E_INVALID if a minus sign is present. I've also added tests in test-misc.c to verify this change along with other basic formatting and overflow edge cases.

### DIFF
--- a/src/util/string.c
+++ b/src/util/string.c
@@ -5,6 +5,7 @@
  * String Helpers
  */
 
+#include <ctype.h>
 #include <c-stdaux.h>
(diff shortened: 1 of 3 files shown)


dvdhrm:

PR #455, opened 2026-07-04
util/string: reject negative numbers in util_strtou32/64

SABITHSAHEB:
While auditing the string parsing utilities in src/util/string.c, I noticed that util_strtou32 and util_strtou64 use strtoul/strtoull directly without validating whether the input string has a negative sign.

Because of standard C library behavior, passing negative values (e.g., "-1" or "-4294967295") parses successfully as very large unsigned integers due to silent modulo arithmetic conversion. In resource limit configurations or parsing contexts, this behavior can allow users to inadvertently bypass limits.

To fix this, I added a lightweight string_is_negative() check at the beginning of the unsigned parsing helpers to reject strictly negative values while continuing to allow valid negative zero (-0) variations. I also added a corresponding test case in test-misc.c to verify this validation.

### DIFF
--- a/src/util/string.c
+++ b/src/util/string.c
@@ -11,11 +11,26 @@
 #include "util/error.h"
 #include "util/string.h"
 
+static bool string_is_negative(const char *string) {
+        while (*string == ' ' || (*string >= '\t' && *string <= '\r'))
+                string++;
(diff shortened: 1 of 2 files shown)


dvdhrm:

PR #456, opened 2026-07-05
util/string: harden integer parsing against signed wrapping and white…

SABITHSAHEB:
While reviewing the utility string parsers, I found they wrapped standard library functions (strtoul / strtol) which accept negative sign prefixes (e.g., -1) and skip whitespaces. This allowed negative inputs to wrap silently into large unsigned integers and bypass range checks.

I replaced these library calls with a strict, single-pass inline digit validation and conversion loop, and added verification test cases in test-misc.c.

### DIFF
--- a/src/util/misc.c
+++ b/src/util/misc.c
@@ -74,7 +74,7 @@ int misc_memfd(const char *name, unsigned int uflags, unsigned int useals) {
          */
         if (!(flags & MISC_MFD_ALLOW_SEALING)) {
                 if (seals & ~MISC_F_SEAL_SEAL)
-                        return error_origin(-ENOTRECOVERABLE);
+                        return -ENOTRECOVERABLE;
 
                 /* Already set by the kernel if sealing is disabled. */
                 seals &= ~MISC_F_SEAL_SEAL;
@@ -86,7 +86,7 @@ int misc_memfd(const char *name, unsigned int uflags, unsigned int useals) {
          * our codebase. For older kernels, we strip these flags automatically.
          */
         if (!(flags & (MISC_MFD_EXEC | MISC_MFD_NOEXEC_SEAL)))
-                return error_origin(-ENOTRECOVERABLE);
+                return -ENOTRECOVERABLE;
 
         /*
          * Create the memfd. If the flags are not supported, strip the EXEC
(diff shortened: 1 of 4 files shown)


src/util/misc.c
```diff
@@ -74,7 +74,7 @@ int misc_memfd(const char *name, unsigned int uflags, unsigned int useals) {
          */
         if (!(flags & MISC_MFD_ALLOW_SEALING)) {
                 if (seals & ~MISC_F_SEAL_SEAL)
-                        return error_origin(-ENOTRECOVERABLE);
+                        return -ENOTRECOVERABLE;
```
dvdhrm, line 77: Why do you remove the `error_origin()` call?

src/util/misc.c
```diff
@@ -86,7 +86,7 @@ int misc_memfd(const char *name, unsigned int uflags, unsigned int useals) {
          * our codebase. For older kernels, we strip these flags automatically.
          */
         if (!(flags & (MISC_MFD_EXEC | MISC_MFD_NOEXEC_SEAL)))
-                return error_origin(-ENOTRECOVERABLE);
+                return -ENOTRECOVERABLE;
```
dvdhrm, line 89: Why do you remove the `error_origin()` call?

src/util/sockopt.c
```diff
@@ -11,6 +11,10 @@
 #include "util/error.h"
 #include "util/sockopt.h"
 
+#ifndef SO_PEERPIDFD
+#  define SO_PEERPIDFD 77
+#endif
+
```
dvdhrm, line 17: Can you elaborate why you added this?

dvdhrm (commented):
You opened 3 PRs changing the same thing. There are several unrelated changes. You repeatedly showed this same behavior with other PRs.

I do not appreciate this behavior. It takes a lot of effort to review such changes, and your PRs show little effort from your side to elaborate _why_ these changes are needed, how the different call-sites are affected, and why you open many different conflicting PRs with just different approaches for the same thing.

Please stop.

SABITHSAHEB:
Understood, and sorry for the noise. To answer the inline questions: the error_origin() removals, the SO_PEERPIDFD define and the empty .wraplock were leftovers from other local branches that shouldn't have been in this diff at all. Opening several overlapping PRs instead of iterating on one was my mistake too. I'll stop.

PR #457, opened 2026-07-11
log: tolerate unavailable journal in lossy mode; launch: always exit on broker death

Sinity:
## Summary

Two robustness fixes: a lost diagnostic must never take down the message bus, and a dead broker must never leave its service masquerading as alive.

> **v2:** the original description attributed the failure to a race with journald's shutdown teardown. Deeper forensics showed that was wrong — the journal was alive throughout; the actual mechanism is socket orphaning, below. Commit messages and in-code comments reworded accordingly, and the new test's `meson.build` registration moved to its alphabetical slot; no functional change.

## Mechanism

A full journald restart — socket units included, which stopping journald for journal-file maintenance requires, since socket activation otherwise resurrects it — destroys and recreates `/run/systemd/journal/socket`. Every already-connected `SOCK_DGRAM` client is permanently orphaned: the kernel fails its next send with `ECONNREFUSED` and every later one with `ENOTCONN`. dbus-broker connects its log socket once at startup and never reconnects.

`log_journal_send()` treats anything but `EAGAIN` as a hard error even in lossy mode, and log errors fold into the dispatch path's return value — so the first lossy diagnostic after such a restart kills the broker (commit 1). The launcher then loses its own SIGCHLD diagnostic to the same orphaned socket and returns before `sd_event_exit()`, so it survives its dead child and the unit keeps counting as running while `dbus.socket` keeps accepting (commit 2).

Observed twice in production (dbus-broker 37, systemd 260.2, NixOS): journald restarted for maintenance; 25 hours later in one case and 4 days in the other, the first lossy diagnostic — a policy send-denial emitted during session teardown at shutdown — killed the broker under a second into the shutdown, the launcher kept the unit alive, and PID 1 spent ~15 minutes in 90 s bus-reconnect timeouts. (That amplification is fixed on the systemd side in systemd/systemd#42981, but deployed systemds retain it.) The journal recorded the broker's own error trace, via the stderr stream, at the very instant the native send failed — journald was alive and healthy both times, so no shutdown ordering is violated or would help.

Scope notes: dropping (rather than reconnecting) is the minimal fix — lossy mode's documented contract already permits it, and drop semantics are needed anyway for windows where no journal endpoint exists; reconnect-on-failure would be a reasonable follow-up on top. `log_stream_send()` has an analogous non-EAGAIN escalation in lossy mode; left unchanged here as it is not the observed path — can follow up if wanted.

## Reproducer

Self-contained; does not touch the host's journald or system bus (private mount namespace, fake journal socket, private socket-activated bus). Run as root, with `dbus-broker-launch` from `$PATH` or as `$1`. Requires `unshare`, `systemd-socket-activate`, `busctl`, `dbus-test-tool`, `setpriv`, `timeout`, `python3`.

<details><summary>reproduce.sh</summary>

```bash
#!/usr/bin/env bash
# Reproducer: dbus-broker exits when the peer of its connected journal
# socket dies (e.g. after a full journald restart including its socket
# units, which orphans every already-connected client).
#
# Fully isolated: a fake journal socket is bind-mounted over
# /run/systemd/journal/socket in a private mount namespace; a private,
# socket-activated bus instance runs against it. The host's journald
# and system bus are not touched.
#
# Usage (as root):  ./reproduce.sh [path/to/dbus-broker-launch]
# Requires: unshare, systemd-socket-activate, busctl, dbus-test-tool,
#           setpriv, timeout, python3
set -eu

LAUNCHER=${1:-$(command -v dbus-broker-launch)}
[ "$(id -u)" = 0 ] || { echo "run as root" >&2; exit 1; }
D=$(mktemp -d /tmp/dbx.XXXXXX)
chmod 755 "$D"
BUS="unix:path=$D/bus.sock"
PIDS=()
trap 'kill ${PIDS[@]+"${PIDS[@]}"} 2>/dev/null || true; rm -rf "$D"' EXIT

cat > "$D/system.conf" <<EOF
<!DOCTYPE busconfig PUBLIC "-//freedesktop//DTD D-BUS Bus Configuration 1.0//EN"
 "http://www.freedesktop.org/standards/dbus/1.0/busconfig.dtd">
<busconfig>
  <type>system</type>
  <auth>EXTERNAL</auth>
  <listen>unix:path=$D/unused.sock</listen>
  <policy user="root">
    <allow own="com.test.Target"/>
    <allow send_destination="*"/>
    <allow receive_sender="*"/>
  </policy>
  <policy context="default">
    <allow user="*"/>
    <allow send_destination="org.freedesktop.DBus"/>
    <deny send_destination="com.test.Target"/>
    <allow receive_sender="*"/>
  </policy>
</busconfig>
EOF

# Fake journald: a SOCK_DGRAM listener standing in for the real journal
# socket. Killing it reproduces what a journald restart with socket
# re-creation does to already-connected clients: their peer dies.
python3 - "$D/journal.socket" > "$D/journal.log" 2>&1 <<'EOF' &
import socket, sys, os
s = socket.socket(socket.AF_UNIX, socket.SOCK_DGRAM)
s.bind(sys.argv[1]); os.chmod(sys.argv[1], 0o666)
print('READY', flush=True)
while True:
    for line in s.recv(65536).split(b'\n'):
        if line.startswith(b'MESSAGE='):
            print(line.decode(errors='replace'), flush=True)
EOF
JPID=$!; PIDS+=("$JPID")
for _ in $(seq 50); do grep -q READY "$D/journal.log" 2>/dev/null && break; sleep 0.1; done

unshare -m bash -c "
  mount --bind '$D/journal.socket' /run/systemd/journal/socket &&
  exec systemd-socket-activate -l '$D/bus.sock' \
    '$LAUNCHER' --scope system --config-file '$D/system.conf'
" > "$D/broker.log" 2>&1 &
PIDS+=("$!")
sleep 1

DBUS_SYSTEM_BUS_ADDRESS=$BUS dbus-test-tool echo --system --name=com.test.Target &
PIDS+=("$!")
sleep 1
chmod 666 "$D/bus.sock"

# The trigger: any diagnostic the broker sends to the journal. A policy
# send-denial is convenient and side-effect-free.
deny() {
  timeout 5 setpriv --reuid=65534 --regid=65534 --clear-groups \
    busctl --address="$BUS" call com.test.Target /x com.test.I Method 2>&1
}

echo "== baseline denial (journal peer alive):"
OUT=$(deny || true); echo "$OUT"
[[ $OUT == *"Access denied"* ]] || { echo "SETUP FAILURE: baseline call did not hit the policy denial" >&2; exit 2; }
sleep 0.5
grep -q 'A security policy denied' "$D/journal.log" \
  || { echo "SETUP FAILURE: diagnostic never reached the journal socket" >&2; exit 2; }
echo "== diagnostic delivered to the journal socket"

kill -9 "$JPID"; wait "$JPID" 2>/dev/null || true      # the journal peer dies; connected clients are orphaned
sleep 0.5
echo "== same denial with dead journal peer:"
OUT2=$(deny || true); echo "$OUT2"
sleep 1

echo "== bus state:"
if [[ $OUT2 == *"Access denied"* ]] \
   && timeout 5 busctl --address="$BUS" call org.freedesktop.DBus \
        /org/freedesktop/DBus org.freedesktop.DBus ListNames >/dev/null 2>&1; then
  echo "PASS: broker survived the lost diagnostic; bus functional"
else
  echo "FAIL: bus dead - the broker exited over a lost diagnostic:"
  grep -m1 -A9 'ERROR log_journal_send' "$D/broker.log" || cat "$D/broker.log"
fi
```

</details>

Unpatched, the second denial kills the broker with the production stack (`log_journal_send +402: Connection refused` → `peer_queue_unicast` → … → `broker_run`) and the launcher survives its dead child — the bus socket keeps accepting, nothing ever answers. Patched, the caller still gets `Access denied`, the diagnostic is silently dropped, and the bus stays fully functional.

## Testing

- New unit test (`src/util/test-log.c`): dead-peer case (`ECONNREFUSED`, then `ENOTCONN`) and pre-existing full-queue `EAGAIN` case. SIGABRTs on unpatched `main`, passes with this PR.
- Reproducer above, run against an unpatched and a patched build (output as described).
- `meson test --suite unit`: 28/29; the `test-fs` failure is pre-existing and environmental (#458).
- Both patches have been in production since 2026-07-15; the two clean shutdowns since then completed in 18 s and 4.7 s.


Sinity:
The PID 1 side of this incident (synchronous GetId blocking the manager for BUS_AUTH_TIMEOUT per reconnection attempt while dbus.socket accepts for a dead broker) is now up as systemd/systemd#42981.

dvdhrm:
Thanks for the PR!

> Two robustness fixes for the same production incident: **an unavailable journal must never take down the message bus, and a dead broker must never leave its service masquerading as alive.**

Why? We depend on the journal (if it is selected), so D-Bus should shutdown before the journal does.

Sinity:
Fair challenge — and the ordering you describe did hold. The "raced journald teardown" framing in v1 was wrong; I've pushed a v2 rewording both commit messages and the PR description to match what actually happened (no functional change; the in-code comments carried the same stale framing and were updated with it).

journald was alive through both captured occurrences — it ingested the broker's own error trace via the stderr stream at the instant the native send failed, and kept recording for the following 15 minutes. The real mechanism: journald had been fully restarted 25 hours earlier in one occurrence and four days in the other (socket units included — required to actually stop it, since socket activation otherwise resurrects it). That destroys and recreates the listening socket, permanently attaching an already-connected `SOCK_DGRAM` client to the dead peer: `ECONNREFUSED` once, `ENOTCONN` ever after (confirmed against kernel source — `unix_dgram_sendmsg()` clears the peer reference on first send failure, with no implicit reconnect). The broker's next lossy diagnostic then took the bus down. Ordering can't fix a failure armed days in advance. The PR description now carries a self-contained reproducer; for comparison, systemd's own components treat the same class of condition — an unreachable journal — as ignorable ("Failed to connect stdout to the journal socket, ignoring: Connection refused").

Lossy mode already drops on `EAGAIN`; a dead peer is the same statement from the kernel with a different spelling. And the launch fix is independent of any of this: a failed diagnostic commit must not cancel `sd_event_exit()`.

dvdhrm:
Can you please write this in your own words? I do not appreciate talking to generated content.

Sinity:
I could plagiarize Fable's output but I fail to understand how is doing so helpful. [Here's](https://claude.ai/code/artifact/478eb71b-f7f7-4afb-94ae-9c2f9061d921) the context regarding extent of my involvement, for transparency. It's not like I told the agent to spam PRs not caring about their quality. If there's a flaw in the PR, it's on me regardless.

dvdhrm:
I do not appreciate being sent generated messages, especially undisclosed. I believe this is very rude. Please refrain from doing so in the future, or find someone else to review and merge this.

I am closing this PR, as I do not see this going forward. If you see a bug you don't know how to fix, please file an issue and describe it. I am not interested in generated patches.

PR #462, opened 2026-08-10
Various bug fixes, policy enforcement hardening, and memory leak cleanups

wltu:
This patch series addresses several edge-case bugs, resource leaks, and policy enforcement issues identified during codebase review and fuzzing of dbus-broker:

### Policy & Access Control
- **`bus/policy`**: Only evaluate `own_verdict` on exact name matches (`!*end`). Exact `<allow own="..."/>` rules previously matched prefix sub-hierarchies erroneously.
- **`bus/driver`**: Check send policy before queueing messages for service auto-activation, preventing unauthorized senders from triggering daemon startup.
- **`launch/config`**: Mark `<allow>`/`<deny>` rules with unresolvable user or group as invalid to avoid dropping identity constraints and failing open during policy import.
- **`launch/launcher`**: Fail startup (`-EINVAL`) if a configured `<user>` cannot be resolved rather than silently continuing execution as root.

### Resource & Descriptor Management
- **`dbus/socket`**: Close ancillary file descriptors in `socket_recvmsg()` if `recvmsg()` encounters a 0-byte EOF or error, preventing descriptor leaks.
- **`bus/listener`**: Gracefully handle transient `EMFILE`, `ENFILE`, `ENOBUFS`, and `ENOMEM` errors in `accept4()` rather than terminating the daemon.
- **`bus/listener`**: Safely unlink active peers from `peer_list` on `listener_deinit()` to avoid touching deallocated listener memory on subsequent peer disconnects.

### Memory Leaks & Assertions
- **`bus/reply`**: Free `ReplySlot` when `user_charge()` fails in `reply_slot_new()`.
- **`broker/controller`**: Free `ControllerReload` when `user_charge()` fails in `controller_reload_new()`.
- **`launch/nss-cache`**: Postorder traverse and free unlinked negative entries from `uid_tree` and `gid_tree` on `nss_cache_deinit()`.
- **`launch/launcher`**: Add `+ 1` null-terminator headroom to `str_max_*` buffers in `launcher_run_child()` to prevent `snprintf` length assertion failures.
- **`launch/policy`**: Skip duplicate console UIDs in `policy_export()` to prevent underflow in `at_console_uids[i] - next_uid`.

### Protocol & Message Handling
- **`dbus/message`**: Check `c_dvar_read()` return values for all header fields in `message_parse_metadata()` to prevent NULL pointer dereferences on malformed variant signatures.
- **`dbus/message`**: Reserve patch buffer headroom in `message_new_incoming()` so synthetic sender stitching never causes `message->n_data` to exceed `MESSAGE_SIZE_MAX`.
- **`util/log`**: Ensure lossy stream logging drops messages on `EAGAIN` / partial send rather than falling through to synchronous blocking I/O.

---

 ### Verification
All changes build cleanly and pass the complete test suite:
- `ninja -C build`
- `meson test -C build` (54 / 54 tests passing)


src/util/log.c
```diff
@@ -366,6 +366,7 @@ static int log_stream_send(Log *log) {
                 n_blob -= l;
 
                 log_warn = true;
+                goto out;
```
dvdhrm, line 369: The introductory comment in this function elaborates why the synchronous send is used in this scenario. Your change possibly discards a partial message, causing any subsequent message to be discarded as well.

Can you elaborate why you think this change is needed?

src/bus/driver.c
```diff
@@ -2600,6 +2600,28 @@ static int driver_forward_unicast(Peer *sender, const char *destination, Message
...
+                );
+                if (r) {
+                        if (r == POLICY_E_ACCESS_DENIED ||
+                            r == POLICY_E_SELINUX_ACCESS_DENIED ||
+                            r == POLICY_E_APPARMOR_ACCESS_DENIED)
+                                return DRIVER_E_SEND_DENIED;
+
+                        return error_fold(r);
+                }
+
```
dvdhrm, line 2624: Unauthorized senders are certainly allowed to trigger service activation. The bus even provides `StartServiceByName()` for exactly that purpose. Hence, I do not understand why this policy check is needed.

Furthermore, you pass `NULL` as seclabel and nameset of the receiver, thus you will skip any policy-checks that match on the seclabel or nameset of the receiver. Can you explain your thought process behind this? I do not understand how this is supposed to work.

src/bus/listener.c
```diff
@@ -116,7 +116,10 @@ int listener_init_with_fd(Listener *l,
  * listener_deinit() - XXX
  */
 void listener_deinit(Listener *listener) {
-        c_assert(c_list_is_empty(&listener->peer_list));
+        Peer *peer, *peer_safe;
+
+        c_list_for_each_entry_safe(peer, peer_safe, &listener->peer_list, listener_link)
+                c_list_unlink(&peer->listener_link);
```
dvdhrm, line 122: Dynamic listener addition/removal is not hooked up anywhere and is not used at runtime. I don't think peer removal is the correct choice here. If we ever hook up listener management at runtime, I think we need to disconnect peers that connected through this listener.

I don't think there is any immediate need for dynamic listener management, though. But if you need this, I would prefer if this becomes a separate PR with proper testing.

src/dbus/socket.c
```diff
@@ -555,6 +571,7 @@ static int socket_recvmsg(Socket *socket,
                 socket_hangup_input(socket);
                 return SOCKET_E_LOST_INTEREST;
         } else if (_c_unlikely_(l < 0)) {
+                socket_close_cmsg_fds(&msg);
```
dvdhrm, line 574: While `msg` is initialized properly, `msg.msg_control` points to an uninitialized buffer. If the kernel returns a negative error code, there is no guarantee that any part of `msg` was adjusted. This fix will thus read uninitialized ancillary data. What am I missing?

Can you elaborate under which conditions the kernel fills in ancillary data when returning a non-positive integer from `recvmsg()`? I am not aware of any, and wouldn't even know whether it is required to update `msg_controllen` in those cases (though this could be argued in the EOF case).

src/bus/listener.c
```diff
@@ -36,6 +36,13 @@ static int listener_dispatch(DispatchFile *file) {
                          */
                         dispatch_file_clear(&listener->socket_file, EPOLLIN);
                         return 0;
+                } else if (errno == EMFILE ||
+                           errno == ENFILE ||
+                           errno == ENOBUFS ||
+                           errno == ENOMEM ||
+                           errno == ECONNABORTED ||
+                           errno == EINTR) {
+                        return 0;
```
dvdhrm, line 45: Can you elaborate under which circumstances dbus-broker can observe `EMFILE` or `ENFILE`? The accounting/quota system should prevent any unbounded FD allocation, and thus I cannot see how this can be triggered.

Can you elaborate why you check for `EINTR`? This is not a syscall with timing-related arguments, and thus `SA_RESTART` should trigger.

For the last three (ENOBUFS, ENOMEM, ECONNABORTED), can you elaborate how we can trigger those, any why ignoring/discarding the new client is the correct thing to do?

src/dbus/message.c
```diff
@@ -67,7 +67,7 @@ int message_new_incoming(Message **messagep, MessageHeader header) {
         }
 
         n_data = c_align_to(n_header, 8) + n_body;
-        if (n_data > MESSAGE_SIZE_MAX)
+        if (n_data + sizeof(message->patch) > MESSAGE_SIZE_MAX)
```
dvdhrm, line 70: `message->patch` is a maximum limit, not the exact addition. So now we reject messages that could end up below MESSAGE_SIZE_MAX even with stitching.

I don't think `MESSAGE_SIZE_MAX` should necessarily be interpreted as a limit for messages you receive, but only for sending. Otherwise, we will run into weird problems once broadcasts adjust message headers depending on receiver.

This is certainly a good catch and deserves a comment explaining the effects. I currently don't think this needs to be fixed, though. Not entirely sure on this.

dvdhrm (changes requested):
Thanks a lot for the PR! Several small fixes (especially the error-path fixes) that look good and which I will likely cherry-pick over the next few days.

However, I also found some weird changes that I do not understand. See my inline comments. I would love to hear what your intention with those is, and how you encountered them.

Thanks
David

wltu:
> Thanks a lot for the PR! Several small fixes (especially the error-path fixes) that look good and which I will likely cherry-pick over the next few days.
> 
> However, I also found some weird changes that I do not understand. See my inline comments. I would love to hear what your intention with those is, and how you encountered them.
> 
> Thanks David

Hi David,
    
Thank you for taking the time to do a detailed review!
    
These changes came out of a comprehensive security, edge-case, and static analysis audit looking at potential denial-of-service, resource leak, and error-recovery paths.
    
Your points on the inline comments make complete sense. I have drop the unnecessary/controversial ones so we keep only the clean bug fixes and memory leak cleanups.

PR #463, opened 2026-08-14
Add configs for user connections ratelimiting

bluca:
The currently available limits do not allow protecting from a broken user that keeps opening and closing thousands of connections, causing the system to crawl to a halt on low powered hardware.

Add new limit settings to the launcher config file:

connections_rate_limit_per_user_sec
connections_rate_limit_per_user_burst

If enabled, each client is only allowed 'burst' new connections every 'sec' seconds, and gets rejected otherwise.


bluca:
quick python script to test this manually in a VM:

```
#!/usr/bin/env python3
"""Trigger the dbus-broker per-user connection rate-limit.

Enable the limit in the bus config first:

    <limit name="connections_rate_limit_per_user_sec">5</limit>
    <limit name="connections_rate_limit_per_user_burst">5</limit>

Usage:
    python3 test/trigger-rate-limit.py
    python3 test/trigger-rate-limit.py --address unix:path=/run/dbus/system_bus_socket
"""
import argparse
import os
from time import sleep

import pydbus
from gi.repository import GLib


def main():
    p = argparse.ArgumentParser(
        description=__doc__, formatter_class=argparse.RawDescriptionHelpFormatter
    )
    p.add_argument("--address", default=os.environ.get("DBUS_SESSION_BUS_ADDRESS"),
                   help="D-Bus address (default: $DBUS_SESSION_BUS_ADDRESS)")
    p.add_argument("-n", "--count", type=int, default=20,
                   help="number of connection attempts (default: 50)")
    args = p.parse_args()

    if not args.address:
        p.error("no bus address; set DBUS_SESSION_BUS_ADDRESS or pass --address")

    accepted = refused = 0
    for i in range(1, args.count + 1):
        try:
            bus = pydbus.connect(args.address)
        except GLib.GError as e:
            refused += 1
            print(f"attempt {i:3}: REFUSED ({e.message})")
            sleep(5)
            continue

        accepted += 1
        print(f"attempt {i:3}: accepted")
        bus.con.close_sync()

    print(f"\n{accepted} accepted, {refused} refused")
    if not refused:
        print("Rate limit not triggered: enable it and ensure burst < count.")


if __name__ == "__main__":
    main()
```

dvdhrm:
Thanks for writing this up! I have to admit, I thought you intended to just stall connections when they are rate-limited, rather than rejecting them. The problem with disconnecting them is that I do not see how we can find sensible limits then.

We have users running dbus-broker on machines with hundreds of CPUs, and they have boot sequences taking many minutes. On a properly parallelized boot, you can easily have many connections being opened in a small burst, and your patches could cause them to be disconnected (and D-Bus has no sensible way to deal with disconnects so far).

I know, you do not enable the rate-limit by default, but I would assume you want distros to set some sane default configurations. So what would you pick there?

Sadly, we do not have one listener per UID, so we cannot "stall" on accept(). But we could certainly keep accepting connections, but never dispatch them if rate-limited, and only resume them once the rate-limit period is over. Maybe I am missing something, but this sounds significantly less disruptive, and would avoid introducing fault paths that we didn't have before.

What do you think?

bluca:
> I know, you do not enable the rate-limit by default, but I would assume you want distros to set some sane default configurations. So what would you pick there?

Actually, my intention was to never set any defaults anywhere. Specific use cases/deployments/products/users/admins can pick a value that matches their environment and expectations.

> Sadly, we do not have one listener per UID, so we cannot "stall" on accept(). But we could certainly keep accepting connections, but never dispatch them if rate-limited, and only resume them once the rate-limit period is over. Maybe I am missing something, but this sounds significantly less disruptive, and would avoid introducing fault paths that we didn't have before.

The problem is that still leaves the service open to unbounded resource usage growth, which is exactly what I needed to avoid...

Maybe I could rename this to something like `connections_reject_rate_limit_per_user_sec` or so, which leaves the door open to also add something like `connections_stall_rate_limit_per_user_sec` or so, if anyone has a use case for it and wants to implement it? Or maybe instead add another knob like `connections_rate_limit_strategy=reject` which leaves open in the future the possibility to add `stall` later as a different strategy

dvdhrm:
> > I know, you do not enable the rate-limit by default, but I would assume you want distros to set some sane default configurations. So what would you pick there?
> 
> Actually, my intention was to never set any defaults anywhere. Specific use cases/deployments/products/users/admins can pick a value that matches their environment and expectations.

So you don't want distributions to pick a default?

> > Sadly, we do not have one listener per UID, so we cannot "stall" on accept(). But we could certainly keep accepting connections, but never dispatch them if rate-limited, and only resume them once the rate-limit period is over. Maybe I am missing something, but this sounds significantly less disruptive, and would avoid introducing fault paths that we didn't have before.
> 
> The problem is that still leaves the service open to unbounded resource usage growth, which is exactly what I needed to avoid...

Does it really? I see 2 scenarios:

1) In a loop, an application creates a connection synchronously (SASL+Hello), dumps a possibly asynchronous call and disconnects. This application will be stalled on the rate-limited synchronous SASL and thus no longer have any unbounded calls.

2) In a loop, an application creates a connection asynchronously, with SASL+Hello+Message+close queued in the pipeline. This will never be affected by the rate-limits, since no synchronous calls are happening and kernel buffers are usually sufficiently large. However, this will hit the fd-limit quota of the user, once more and more connections are rate-limited and thus pending in the broker. So at some point, new connections of the user will be immediately rejected.

I have never seen anyone but dbus-broker to do SASL+Hello+Message in an asynchronous fashion, so I don't think 2) is a concern. But even if it was, it would lead to the same behavior as your proposed patch, but via the fd-quota rather than a new quota.

I really think 1) is the more realistic issue, and this would be caught by stalling SASL (or any kind of dispatch).

A misconfiguration of the rate-limit would no longer lead to disconnects, but only a delayed processing of messages. With suitable logging, this sounds like the more graceful way to handle it, doesn't it?

Am I missing something?

